PR #5, opened 2026-06-08
feat(agents): Add verifier agent package

yuwenma:
This PR adds the verification engine. It provides a structured framework used by scenario managers and task evaluators (will be added separately) to validate cluster state (e.g., workload readiness, replica convergence) during and after chaos disruptions or operational tasks (will be added separately).

### DIFF
--- a/devops_bench/agents/verifier/spec.py
+++ b/devops_bench/agents/verifier/spec.py
@@ -1,0 +1,27 @@
+# Copyright 2026 The Kubernetes Authors.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+from typing import Dict, Union
+from pydantic import RootModel
+from devops_bench.agents.verifier.verifiers.pod_healthy_verifier import PodHealthyVerifier
+from devops_bench.agents.verifier.verifiers.scaling_complete_verifier import ScalingCompleteVerifier
+
+# SingleVerificationSpec is a discriminated union of all supported checker types
+SingleVerificationSpec = Union[PodHealthyVerifier, ScalingCompleteVerifier]
+
+# Top-level VerificationSpec which parses a dict mapping spec IDs to single verification specs.
+class VerificationSpec(RootModel[Dict[str, SingleVerificationSpec]]):
+    """Represents a structured verification specification."""
+    pass
+
--- a/devops_bench/agents/verifier/test_verifier.py
+++ b/devops_bench/agents/verifier/test_verifier.py
@@ -1,0 +1,216 @@
+# Copyright 2026 The Kubernetes Authors.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import json
+import subprocess
+from unittest.mock import MagicMock, patch
+
+import pytest
+
+from devops_bench.agents.verifier.verifier import VerifierAgent
+from devops_bench.agents.verifier.verifiers.pod_healthy_verifier import PodHealthyVerifier
+from devops_bench.agents.verifier.verifiers.scaling_complete_verifier import ScalingCompleteVerifier
+
+@pytest.fixture(autouse=True)
+def fake_time():
+    """Fixture to mock time.perf_counter and time.sleep globally to avoid real-time waiting/busy loops."""
+    current_time = [0.0]
+
+    def mock_perf_counter():
+        current_time[0] += 0.001
+        return current_time[0]
+
+    def mock_sleep(seconds):
+        current_time[0] += seconds
+
+    with patch("time.perf_counter", side_effect=mock_perf_counter), \
+         patch("time.sleep", side_effect=mock_sleep):
+        yield
+
+@pytest.fixture
+def verifier_agent():
+    return VerifierAgent()
+
+@patch("subprocess.run")
+def test_pod_healthy_verifier_check_status_success(mock_run):
+    mock_output = json.dumps(
+        {
+            "items": [
+                {
+                    "status": {
+                        "phase": "Running",
+                        "conditions": [{"type": "Ready", "status": "True"}]
+                    }
+                },
+                {
+                    "status": {
+                        "phase": "Running",
+                        "conditions": [{"type": "Ready", "status": "True"}]
+                    }
+                },
+            ]
+        }
+    )
+    mock_run.return_value = MagicMock(
+        stdout=mock_output, returncode=0
+    )
+
+    p_verifier = PodHealthyVerifier(selector="app=my-app")
+    details = p_verifier._get_pods_details()
+    success = p_verifier._check_pods_status(details)
+
+    assert success is True
+    assert len(details["items"]) == 2
+
+@patch("subprocess.run")
+def test_pod_healthy_verifier_check_status_failure(mock_run):
+    mock_output = json.dumps(
+        {
+            "items": [
+                {
+                    "status": {
+                        "phase": "Running",
+                        "conditions": [{"type": "Ready", "status": "True"}]
+                    }
+                },
+                {
+                    "status": {
+                        "phase": "Pending",
+                        "conditions": [{"type": "Ready", "status": "False"}]
+                    }
+                },
+            ]
+        }
+    )
+    mock_run.return_value = MagicMock(
+        stdout=mock_output, returncode=0
+    )
+
+    p_verifier = PodHealthyVerifier(selector="app=my-app")
+    details = p_verifier._get_pods_details()
+    success = p_verifier._check_pods_status(details)
+
+    assert success is False
+
+@patch("subprocess.run")
+def test_scaling_complete_verifier_check_scaling_success(mock_run):
+    mock_output = json.dumps({"status": {"readyReplicas": 3}})
+    mock_run.return_value = MagicMock(
+        stdout=mock_output, returncode=0
+    )
+
+    s_verifier = ScalingCompleteVerifier(deployment="my-dep", min_replicas=3)
+    success, details = s_verifier._check_scaling()
+
+    assert success is True
+    assert "Ready replicas (3) >= min replicas (3)" in details["reason"]
+
+@patch("subprocess.run")
+def test_pod_healthy_verifier_verify_wait_success(mock_run):
+    mock_run.return_value = MagicMock(
+        stdout="pod/my-pod condition met", returncode=0
+    )
+
+    p_verifier = PodHealthyVerifier(selector="app=my-app")
+    result = p_verifier.verify(timeout_sec=60)
+
+    assert result.success is True
+    assert result.reason == "Condition met via kubectl wait"
+
+@patch("subprocess.run")
+def test_pod_healthy_verifier_verify_wait_failure_fallback_success(mock_run):
+    # Mock wait fails, but get pods status succeeds (running fallback)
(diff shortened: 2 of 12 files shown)


devops_bench/agents/verifier/test_verifier.py
```diff
@@ -0,0 +1,172 @@
...
+import json
+import subprocess
+import unittest
+from unittest.mock import MagicMock, patch
+
+from devops_bench.agents.verifier.verifier import VerifierAgent
+from devops_bench.agents.verifier.pod_healthy import PodHealthyVerifier
+from devops_bench.agents.verifier.scaling_complete import ScalingCompleteVerifier
+
+class TestVerifierAgent(unittest.TestCase):
```
pradeepvrd: We have `pytest` as a dependency and other tests are already written using it. Can we please standardize on using pytest for all tests?

devops_bench/agents/verifier/verifiers/scaling_complete_verifier.py
pradeepvrd, line 1: can we use a consistent file name scheme for all verifiers? `scaling_complete_verifier.py`?

Ideally it would be good to place all verifiers in a subdirectory (named `verifiers`) for better seperation between the verifier agent code and the actual verifiers.

devops_bench/agents/verifier/pod_healthy.py
```diff
@@ -0,0 +1,87 @@
...
+from typing import Literal, Optional
+from devops_bench.agents.verifier.base import BaseVerifier, VerificationResult
+
+class PodHealthyVerifier(BaseVerifier):
+    type: Literal["pod_healthy"] = "pod_healthy"
+    selector: str
+    namespace: Optional[str] = None
+
+    def verify(self, timeout_sec: int) -> VerificationResult:
+        start_time = time.time()
```
pradeepvrd: It is preferable to use `time.perf_counter()` when measuring elapsed time as it is guaranteed to be monotonic (unlike `time.time()`)

Applies to a number of other instances in this PR as well.

devops_bench/agents/verifier/scaling_complete.py
```diff
@@ -0,0 +1,83 @@
...
+            success, details = self._check_scaling()
+            if success:
+                return VerificationResult(
+                    success=True,
+                    elapsed_time=time.time() - start_time,
+                    reason=f"Scaling complete: {details.get('reason')}",
+                    details=details,
+                )
+            time.sleep(delay)
+            delay = min(delay * 2, max_delay)
```
pradeepvrd: Curious why delay is being exponentially increased here.

devops_bench/agents/verifier/pod_healthy.py
```diff
@@ -0,0 +1,87 @@
...
+                elapsed_time=time.time() - start_time,
+                reason="Condition met via kubectl wait",
+                details={"output": result.stdout.strip()},
+            )
+        except subprocess.CalledProcessError as e:
+            # Fallback or get details on failure
+            elapsed = time.time() - start_time
+            details = self._get_pods_details()
+            success = self._check_pods_status(details)
+            if success:
```
pradeepvrd: Should we still return success if the subprocess errored out?

devops_bench/agents/verifier/scaling_complete.py
```diff
@@ -0,0 +1,83 @@
...
+        start_time = time.time()
+        delay = 1
+        max_delay = 10
+
+        while time.time() - start_time < timeout_sec:
+            success, details = self._check_scaling()
+            if success:
+                return VerificationResult(
+                    success=True,
+                    elapsed_time=time.time() - start_time,
```
pradeepvrd: If measuring elapsed_time is going to be a common pattern, consider adding a context manager utility that you can reuse. As an example:

```python
import time

class Timer:
    def __enter__(self):
        self.start = time.perf_counter()
        return self

    def __exit__(self, exc_type, exc_val, exc_tb):
        self.elapsed = time.perf_counter() - self.start
```

You can then use it as:

```python
with Timer() as t:
    # Your code here

print(f"Elapsed: {t.elapsed} seconds")
```

devops_bench/agents/verifier/verifier.py
```diff
@@ -0,0 +1,78 @@
...
+        start_time = time.time()
+
+        if isinstance(root, list):
+            results = []
+            overall_success = True
+            overall_reason = []
+            for i, sub_spec in enumerate(root):
+                elapsed = time.time() - start_time
+                remaining_timeout = max(1, timeout_sec - int(elapsed))
+                sub_result = self.wait_for_condition(VerificationSpec(sub_spec), timeout_sec=remaining_timeout)
```
pradeepvrd: Recursively passing in remaining timeout feels like a very brittle way of handling timeout. If you want to have a global timeout for the verification of all specs, the pythonic way would be to add a timeout decorator. It should be fairly straightforward to add one as a utility and can be reused as necessary.

devops_bench/agents/verifier/spec.py
```diff
@@ -0,0 +1,26 @@
...
+from typing import Dict, List, Union
+from pydantic import RootModel
+from devops_bench.agents.verifier.pod_healthy import PodHealthyVerifier
+from devops_bench.agents.verifier.scaling_complete import ScalingCompleteVerifier
+
+# SingleVerificationSpec is a discriminated union of all supported checker types
+SingleVerificationSpec = Union[PodHealthyVerifier, ScalingCompleteVerifier]
+
+# Top-level VerificationSpec which can parse a dict, a list, or a single checker spec.
+class VerificationSpec(RootModel[Union[Dict[str, SingleVerificationSpec], List[SingleVerificationSpec], SingleVerificationSpec]]):
```
pradeepvrd: Wondering if we need to support all these variants. Wouldn't it be simpler to just support `Dict`? We can model the other types as a `dict` with some constraint on the caller and can be much easier to maintain.

yuwenma: One big advantage of K8s API (spec and status) is its strong structured format even in YAML/JSON config. Golang enforces that by defining the structures and built it own k8s/yaml to parse the type. But Python does not have corrsponding type check if simply using the built in json parser lib. So I apply a structured type to enforce the same level of K8s YAML validaiton.

devops_bench/agents/verifier/verifier.py
yuwenma: Right, this is why we need the event watcher. This PR purely migrates what's already agreed in gke-labs. I have finished the event-watcher PR, which needs this one to go in first.

devops_bench/agents/verifier/scaling_complete.py
yuwenma: I like it. done

devops_bench/agents/verifier/scaling_complete.py
yuwenma: Because k8s uses exponential backoff as well. This is a simple exponential retry. Eventually, I think we can build a k8s  controller for this.

devops_bench/agents/verifier/pod_healthy.py
yuwenma: sg

devops_bench/agents/verifier/test_verifier.py
yuwenma: Sg. Done

janetkuo:
@yuwenma are the Copilot review comments addressed? If so, would you resolve them to make it clear?

yuwenma:
> @yuwenma are the Copilot review comments addressed? If so, would you resolve them to make it clear?

They are all resolved and marked now.

devops_bench/agents/verifier/verifiers/pod_healthy_verifier.py
```diff
@@ -0,0 +1,112 @@
...
+                    success=True,
+                    elapsed_time=timer.elapsed,
+                    reason="Condition met via kubectl wait",
+                    details={"output": result.stdout.strip()},
+                )
+            except (subprocess.CalledProcessError, subprocess.TimeoutExpired) as e:
+                # Catch timeout or error and fallback to polling check
+                # Check status details
+                details = self._get_pods_details(timeout=min(KUBECTL_DEFAULT_TIMEOUT, remaining))
+                last_details = details
```
janetkuo, line 66: When `remaining` is very small (e.g., `0.01` seconds) near the end of the timeout budget, `kubectl get pods` will fail unconditionally. 

Consider doing something similar as `ScalingCompleteVerifier` (which uses `timeout=max(1.0, remaining)`) for the fallback check, and adding an explicit final check after the loop to ensure a reliable final evaluation.

devops_bench/agents/verifier/verifiers/scaling_complete_verifier.py
```diff
@@ -0,0 +1,92 @@
...
+            if self.namespace:
+                cmd.extend(["-n", self.namespace])
+
+            result = subprocess.run(
+                cmd, capture_output=True, text=True, check=True, timeout=timeout
+            )
+            dep_data = json.loads(result.stdout)
+            ready_replicas = dep_data.get("status", {}).get("readyReplicas", 0)
+            success = ready_replicas >= self.min_replicas
+            reason = (
```
janetkuo, line 90: Verify `status.observedGeneration >= metadata.generation` to be fully robust against race conditions where `kubectl get deployment` is queried before the controller has observed the latest scaling action.

devops_bench/agents/verifier/spec.py
pradeepvrd: My question was from an API perspective, do we want to support something so complex. Can we not define `Dict[str, SingleVerificationSpec]` as the interface? We enforce an id for any spec that we validate and it keeps interface and validation much simpler. Unless there's a strong reason for us to handle all these different data structures as input, I propose we keep the API interface very simple.

devops_bench/agents/verifier/verifier.py
pradeepvrd: SGTM.

pyproject.toml
pradeepvrd, line 1: Can you also lock the dependencies and add uv.lock here?

If you ensure you are enabling the pre-commit hooks as described in the AGENTS.md, it will block you from pushing automatically.

pradeepvrd:
@yuwenma Can you please squash the changes to keep commits atomic.

devops_bench/agents/verifier/spec.py
yuwenma: Good point. If we want to follow the K8s API convension https://github.com/kubernetes/community/blob/main/contributors/devel/sig-architecture/api-conventions.md#spec-and-status, we should define the `Dict[str, SingleVerificationSpec]` because the interface has to be deterministic. If we don't plan to eventually converge to k8s CRD, I'm fine to loose the constraints here.

pradeepvrd: I don't think we aim to converge to k8s CRD. That assumes that k8s event streams are the only source of verification which is not likely to be the case. We might have to verify a variety of conditions for state convergence.

I consider this more akin to `unstructured.Unstructured` which is essentially `map[[string]interface{}`. At this point we do not have clarity on the full API range of the verifier so I suggest we keep it simple and lean on a bit more specification on client side.

janetkuo:
Just noticed https://github.com/kubernetes-sigs/devops-bench/pull/8 depends on this PR. Is there a specific reason why the verifier agent doesn't use kubewatch directly, instead of breaking up into 2 commits/PRs?

yuwenma:
> Just noticed #8 depends on this PR. Is there a specific reason why the verifier agent doesn't use kubewatch directly, instead of breaking up into 2 commits/PRs?

These two PRs separate the pure migration (previously approved in gke-labs) and the new features to avoid dup-review.

devops_bench/agents/verifier/verifiers/scaling_complete_verifier.py
yuwenma: Good point! Thanks

devops_bench/agents/verifier/verifiers/pod_healthy_verifier.py
yuwenma: good catch. Done

devops_bench/agents/verifier/spec.py
yuwenma: Thanks for the clarification. Simplified the verifier Spec.

pyproject.toml
yuwenma: Woohoo, +1 to AGENTS.md.

Done

janetkuo (changes requested):
> These two PRs separate the pure migration (previously approved in gke-labs) and the new features to avoid dup-review.

@yuwenma let's just keep PR #8 (with 2 commits) and squash at merge time (I've applied the squash-merge label on PR #8).

If they're merged as 2 distinct commits, the main branch history will contain an intermediate commit with legacy kubectl wait polling loops that are immediately deleted in the next commit. The main branch should only reflect the final, working architectural state.

Would you rebase PR #8 to resolve the current merge conflicts with main, ensure the review feedback from this PR is carried over, and we move the review there?

janetkuo:
As suggested above, let's move the review / discussion to #8.

PR #18, opened 2026-07-08
chore: pin ruff exactly and format the tree

pradeepvrd:
Pin ruff exactly (==0.15.17) so a lock refresh cannot shift formatter style, fix the two outstanding lints, and format agents/ + chaos/ so the CI gate starts green.

CI itself now lives in Prow rather than a GitHub Actions workflow (review feedback): the presubmit job config is proposed in kubernetes/test-infra#37457 (ruff lint + format check, unit tests on py3.12-3.14, pytest run from the repo root so in-package *_test.py files are collected).


pradeepvrd:
@coderabbit review

.github/workflows/guardrails.yml
```diff
@@ -0,0 +1,44 @@
...
+      - name: Lint
+        run: uv run ruff check .
+
+      - name: Format
+        # ruff is pinned exactly (==) in pyproject so this check cannot drift
+        # with formatter style changes across ruff releases.
+        run: uv run ruff format --check .
+
+      - name: Unit tests
+        run: uv run pytest tests/unit -q
```
janetkuo: Restricting `pytest` to `tests/unit` skips test files located in package directories (such as `devops_bench/agents/chaos/chaos_test.py`)

.github/workflows/guardrails.yml
```diff
@@ -0,0 +1,44 @@
+# CI guardrail: uv sync, ruff lint + format, unit tests.
```
janetkuo: SIG repos should use Prow for CI, so that the integration is better (prow bots respond to commands, blocks merges, etc.)

ref https://github.com/kubernetes/test-infra/tree/master/config/jobs/kubernetes-sigs/agent-sandbox

.github/workflows/guardrails.yml
pradeepvrd: An upcoming PR will move those tests under the `tests/unit`  as well. So this will be a very short lived gap.

.github/workflows/guardrails.yml
pradeepvrd: Ack. I can remove the github workflows and  send a PR to test-infra repo instead.

pradeepvrd (commented):
Addressed comments

janetkuo (approved):
/lgtm

PR #20, opened 2026-07-09
feat(core): k8s wrapper classes for kubectl and conditions

eugeneng04:
This PR introduces thin, shell-free wrappers around the `kubectl` command line and a
  generic wait/poll engine for Kubernetes resources.
It includes:
    - `devops_bench/k8s/kubectl.py`: Wrappers for standard `kubectl` operations (`apply`,
  `get`, `wait`, `rollout status`, and `port-forward`).
    - `devops_bench/k8s/conditions.py`: A generic `poll_until` engine with exponential backoff
  for checking conditions.
    - Co-located unit tests under `tests/unit/k8s/`.
    - A minor fix to `hack/boilerplate.py` to exclude `third_party/` directories from the
  license header check.

## Summary by CodeRabbit

* **New Features**
  * Added a unified Kubernetes helper interface for applying manifests, fetching resources, waiting on readiness/conditions, checking rollout status, and running port-forward.
  * Introduced a polling utility with configurable timeouts and exponential backoff to reliably wait for Kubernetes state changes.
* **Tests**
  * Expanded unit test coverage for command/argument construction, JSON parsing, timeout/backoff behavior, and port-forward lifecycle (including termination/escalation handling).
* **Chores**
  * Updated generated header exclusions to ignore `third_party/**` content.

### DIFF
--- a/devops_bench/k8s/kubectl.py
+++ b/devops_bench/k8s/kubectl.py
@@ -1,0 +1,298 @@
+# Copyright 2026 The Kubernetes Authors.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""Thin, shell-free wrappers around the ``kubectl`` command line."""
+
+from __future__ import annotations
+
+import contextlib
+import json
+import os
+import subprocess
+import time
+from collections.abc import Iterator
+from typing import Any, Protocol
+
+from devops_bench.core import get_logger
+from devops_bench.core.subprocess import CompletedProcess, run
+
+__all__ = [
+    "apply",
+    "get_resource",
+    "port_forward",
+    "rollout_status",
+    "wait",
+]
+
+_log = get_logger("k8s.kubectl")
+
+# Seconds to let ``kubectl port-forward`` establish the tunnel before yielding.
+_PORT_FORWARD_SETTLE_SEC = 3
+
+
+class KubeconfigProvider(Protocol):
+    """Structural type for objects that expose a kubeconfig path."""
+
+    kubeconfig_path: str | None
+
+
+# Accepts a raw kubeconfig path or any object exposing ``kubeconfig_path``.
+type KubeconfigSource = str | KubeconfigProvider | None
+
+
+def _resolve_kubeconfig(kubeconfig: KubeconfigSource) -> str | None:
+    """Return a kubeconfig path from a string or a provider object.
+
+    Args:
+        kubeconfig: Explicit path, a ``KubeconfigProvider``, or None.
+
+    Returns:
+        The kubeconfig path, or None when none is available.
+    """
+    if kubeconfig is None or isinstance(kubeconfig, str):
+        return kubeconfig
+    return kubeconfig.kubeconfig_path
+
+
+def _namespace_args(namespace: str | None) -> list[str]:
+    return ["-n", namespace] if namespace else []
+
+
+def _selector_args(selector: str | None) -> list[str]:
+    return ["-l", selector] if selector else []
+
+
+def _run_kubectl(argv: list[str], kubeconfig: KubeconfigSource, **kwargs: Any) -> CompletedProcess:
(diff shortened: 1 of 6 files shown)


devops_bench/k8s/kubectl.py
```diff
@@ -0,0 +1,298 @@
...
+    remote = remote_port if remote_port is not None else local_port
+    argv = [
+        "kubectl",
+        "port-forward",
+        target,
+        f"{local_port}:{remote}",
+        *_namespace_args(namespace),
+    ]
+    path = _resolve_kubeconfig(kubeconfig)
+    env = {**os.environ, "KUBECONFIG": path} if path else None
```
janetkuo: Consider reusing the shared environment logic from `devops_bench.core.subprocess` instead of directly constructing `{**os.environ, "KUBECONFIG": path}` to ensure consistent environment handling.

devops_bench/k8s/kubectl.py
```diff
@@ -0,0 +1,298 @@
...
+        except Exception as exc:  # noqa: BLE001 - never mask the raise reason
+            _log.warning("error reaping early-exited port-forward: %s", exc)
+        raise RuntimeError(f"kubectl port-forward exited early (code {returncode}) for {target}")
+
+    try:
+        yield
+    finally:
+        _log.info("terminating port-forward to %s...", target)
+        process.terminate()
+        process.wait()
```
janetkuo: `process.wait()` without a timeout can block indefinitely if the `kubectl port-forward` process ignores `SIGTERM` or hangs. Escalate to `process.kill()` after a timeout.

tests/unit/k8s/test_k8s_kubectl.py
```diff
@@ -0,0 +1,199 @@
...
+from devops_bench.core.errors import SubprocessError
+from devops_bench.k8s import kubectl
+
+
+def _completed(stdout: str = "", returncode: int = 0) -> subprocess.CompletedProcess:
+    """Build a real CompletedProcess for the patched ``run``."""
+    return subprocess.CompletedProcess(args=[], returncode=returncode, stdout=stdout)
+
+
+def test_wait_builds_argv_and_threads_kubeconfig(mocker):
```
janetkuo: Add explicit `-> None` return type

tests/unit/k8s/test_k8s_kubectl.py
```diff
@@ -0,0 +1,199 @@
...
+
+
+def test_wait_propagates_subprocess_error(mocker):
+    mocker.patch(
+        "devops_bench.k8s.kubectl.run",
+        side_effect=SubprocessError(["kubectl", "wait"], returncode=1),
+    )
+
+    with pytest.raises(SubprocessError):
+        kubectl.wait("pod", timeout_sec=10)
```
janetkuo, line 200: Consider adding unit tests for the `port_forward` context manager, testing early process exit before `yield`, normal lifecycle cleanup, and `KUBECONFIG` environment threading.

tests/unit/k8s/test_k8s_conditions.py
```diff
@@ -0,0 +1,141 @@
...
+
+    def monotonic(self) -> float:
+        return self.now
+
+    def sleep(self, seconds: float) -> None:
+        self.sleeps.append(seconds)
+        self.now += seconds
+
+
+def test_poll_until_returns_true_immediately_without_sleeping():
```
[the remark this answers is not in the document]
janetkuo: +1

devops_bench/k8s/kubectl.py
eugeneng04: switched to `_build_env` from `core.subprocess` so the port-forward child env is built the same way as `run()`.

One note: `_build_env` is currently private, so if we expect other modules to reuse it we may want to promote it to a public helper. Happy to do that here or in a follow-up.

devops_bench/k8s/kubectl.py
eugeneng04: `wait()` now has a bounded grace period and escalates to `kill()` on timeout. I also bounded the `wait()` after `kill()` so teardown can't hang even in the rare case a process survives SIGKILL; in that case we log an error and abandon it rather than block the with exit. Let me know if you'd rather it block/raise there instead.

tests/unit/k8s/test_k8s_kubectl.py
eugeneng04: Added to all test functions

tests/unit/k8s/test_k8s_kubectl.py
eugeneng04: Added coverage for the `port_forward` lifecycle: argv construction, explicit vs defaulted remote port, KUBECONFIG env threading, early-exit `RuntimeError`, SIGTERM→SIGKILL escalation, and the survives-SIGKILL abandon path.

janetkuo (approved):
/lgtm

PR #27, opened 2026-07-16
Add the result reporter

jessie1111101:
Adds `devops_bench/evalharness/reporter.py` — a `ResultReporter` that mints a
unique, filesystem-safe per-run directory and writes the run's artifacts
(`results.json`, `rows.json`, `manifest.json`) verbatim, so the payload it's
handed is never reshaped.

Unit tests cover the three write helpers and run-directory creation
(uniqueness, `last_run_dir`, run-id suffixing, and run-id sanitization).

/kind feature

```release-note
NONE
```

## Summary by CodeRabbit

* **New Features**
  * Added a result reporting utility that creates timestamped run folders under a configurable root.
  * Writes evaluation outputs to separate JSON files (`results.json`, `rows.json`, and `manifest.json`) and returns the generated paths.
  * Supports an optional run identifier to disambiguate concurrent runs using a filesystem-safe directory suffix.
* **Tests**
  * Added unit tests covering run directory creation, JSON persistence, uniqueness, `last_run_dir` tracking, and run ID sanitization.

### DIFF
--- a/devops_bench/evalharness/reporter.py
+++ b/devops_bench/evalharness/reporter.py
@@ -1,0 +1,152 @@
+# Copyright 2026 The Kubernetes Authors.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""Result sink: write the per-run ``results.json`` and collect artifacts.
+
+The harness depends on a :class:`ResultReporter` rather than calling
+:func:`json.dump` inline, so output sinks (DB, dashboard, blob store) can be
+added by substituting the reporter without editing the orchestrator.
+"""
+
+from __future__ import annotations
+
+import datetime
+import json
+import os
+import re
+from pathlib import Path
+from typing import Any
+
+from devops_bench.core import get_logger
+
+__all__ = ["ResultReporter"]
+
+_log = get_logger("evalharness.reporter")
+
+_RESULTS_FILENAME = "results.json"
+_ROWS_FILENAME = "rows.json"
+_MANIFEST_FILENAME = "manifest.json"
+
+# Characters not safe in a run-directory name are replaced with a dash.
+_SANITIZE_RUN_ID_RE = re.compile(r"[^A-Za-z0-9_-]")
+
+
+class ResultReporter:
+    """Default reporter writing ``results.json`` to a timestamped run directory.
+
+    Every value written through this reporter has already been routed through
+    ``MetricScore.to_entry()`` / ``AgentResult.to_dict()`` /
+    ``ChaosResult.model_dump()`` / ``VerificationResult.model_dump()`` by the
+    orchestrator, so the reporter itself only knows how to JSON-encode the
+    final dicts.
+
+    Args:
+        results_root: Root directory under which per-run subdirectories are
+            created. Defaults to ``"results"``.
+        run_dir_prefix: Prefix for the timestamped subdirectory name.
+        run_id: Optional run identifier appended to the run-dir name. The
+            timestamp alone has 1-second resolution, so two processes started
+            in the same second would otherwise collide on one directory;
+            appending the (process-unique) run id keeps concurrent runs apart.
+
+    Attributes:
+        last_run_dir: The most recent directory returned by
+            :meth:`new_run_dir`, or ``None`` if none has been created yet.
+    """
+
+    def __init__(
+        self,
+        results_root: str | os.PathLike[str] = "results",
+        *,
+        run_dir_prefix: str = "run_",
+        run_id: str | None = None,
+    ) -> None:
+        self.results_root = Path(results_root)
+        self.run_dir_prefix = run_dir_prefix
+        self.run_id = run_id
+        self.last_run_dir: Path | None = None
+
+    def new_run_dir(self) -> Path:
+        """Create and return a fresh timestamped run directory under the root.
+
+        The name carries a UTC timestamp. A supplied ``run_id`` is appended
+        (filesystem-sanitized); without one, sub-second precision is added so two
+        runs started in the same second still get distinct directories.
+
+        Returns:
+            The absolute path of the newly created directory.
+        """
+        now = datetime.datetime.now(datetime.UTC)
+        name = f"{self.run_dir_prefix}{now.strftime('%Y%m%d_%H%M%S')}"
+        if self.run_id:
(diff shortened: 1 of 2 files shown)


devops_bench/evalharness/reporter.py
```diff
@@ -0,0 +1,146 @@
...
+
+        When a ``run_id`` was supplied it is appended (filesystem-sanitized) so
+        concurrent runs sharing a start-second do not collide on one directory.
+
+        Returns:
+            The absolute path of the newly created directory.
+        """
+        timestamp = datetime.datetime.now().strftime("%Y%m%d_%H%M%S")
+        name = f"{self.run_dir_prefix}{timestamp}"
+        if self.run_id:
```
itssimrank, line 92: Since run_id is optional, does the "collision-proof" property claimed in the docstring actually hold for the default case?

devops_bench/evalharness/reporter.py
```diff
@@ -0,0 +1,146 @@
...
+    def new_run_dir(self) -> Path:
+        """Create and return a fresh timestamped run directory under the root.
+
+        When a ``run_id`` was supplied it is appended (filesystem-sanitized) so
+        concurrent runs sharing a start-second do not collide on one directory.
+
+        Returns:
+            The absolute path of the newly created directory.
+        """
+        timestamp = datetime.datetime.now().strftime("%Y%m%d_%H%M%S")
```
itssimrank: This uses local time right? Should we standardize to UTC?

janetkuo:
@coderabbitai review

devops_bench/evalharness/reporter.py
```diff
@@ -0,0 +1,146 @@
...
+        name = f"{self.run_dir_prefix}{timestamp}"
+        if self.run_id:
+            name = f"{name}_{re.sub(r'[^A-Za-z0-9_-]', '-', self.run_id)}"
```
janetkuo: Pre-compile at module level (e.g. `_SANITIZE_RUN_ID_RE = re.compile(r"[^A-Za-z0-9_-]")`) to avoid recompiling the regex pattern on every call

devops_bench/evalharness/reporter.py
```diff
@@ -0,0 +1,146 @@
...
+        run_dir = self.results_root / name
+        run_dir.mkdir(parents=True, exist_ok=True)
```
janetkuo: `exist_ok=True` allows runs created within the same second (when `run_id` is omitted) to silently reuse the existing directory and overwrite previous files. Setting `exist_ok=False` (or raising/resolving collisions) ensures runs do not silently overwrite existing output.

devops_bench/evalharness/reporter.py
janetkuo: +1 to use UTC

devops_bench/evalharness/reporter.py
```diff
@@ -0,0 +1,146 @@
...
+        concurrent runs sharing a start-second do not collide on one directory.
+
+        Returns:
+            The absolute path of the newly created directory.
+        """
+        timestamp = datetime.datetime.now().strftime("%Y%m%d_%H%M%S")
+        name = f"{self.run_dir_prefix}{timestamp}"
+        if self.run_id:
+            name = f"{name}_{re.sub(r'[^A-Za-z0-9_-]', '-', self.run_id)}"
+        run_dir = self.results_root / name
```
janetkuo: The docstring says returning the absolute path of the newly created directory. Returning `(self.results_root / name).resolve()` satisfies this contract when `results_root` is passed as a relative path.

devops_bench/evalharness/reporter.py
jessie1111101: Switched to UTC.

devops_bench/evalharness/reporter.py
jessie1111101: Right — it only held when a `run_id` was passed. The default now appends sub-second precision so same-second runs get separate dirs, and I fixed the docstring to say so.

devops_bench/evalharness/reporter.py
```diff
@@ -0,0 +1,146 @@
...
+        Args:
+            run_dir: Run directory previously returned by :meth:`new_run_dir`.
+            results: Per-task result dicts already shaped to the on-disk schema.
+
+        Returns:
+            The path of the JSON file written.
+        """
+        path = Path(run_dir) / _RESULTS_FILENAME
+        with open(path, "w") as f:
+            json.dump(results, f, indent=2)
```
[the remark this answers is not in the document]
jessie1111101: Added `encoding="utf-8"` to the three writes and the three test reads.

devops_bench/evalharness/reporter.py
jessie1111101: Done — hoisted to `_SANITIZE_RUN_ID_RE` at module level.

devops_bench/evalharness/reporter.py
jessie1111101: Went with resolving collisions over `exist_ok=False` alone: the default dir name now carries sub-second precision, so same-second runs no longer share a path, and I set `exist_ok=False` on top so a collision can't silently overwrite. Can switch to a plain hard-fail if you'd rather.

devops_bench/evalharness/reporter.py
jessie1111101: Done.

devops_bench/evalharness/reporter.py
jessie1111101: Now returns `(results_root / name).resolve()`, so it matches the docstring's absolute-path contract.

devops_bench/evalharness/reporter.py
```diff
@@ -0,0 +1,152 @@
...
+        Args:
+            run_dir: Run directory previously returned by :meth:`new_run_dir`.
+            results: Per-task result dicts already shaped to the on-disk schema.
+
+        Returns:
+            The path of the JSON file written.
+        """
+        path = Path(run_dir) / _RESULTS_FILENAME
+        with open(path, "w", encoding="utf-8") as f:
+            json.dump(results, f, indent=2)
```
janetkuo, line 113: write, write_rows, and write_manifest repeat the same open/json.dump/log/return body. Factor the shared logic into a private helper, e.g. _write_json(). Not a blocker.

janetkuo (approved):
/lgtm

devops_bench/evalharness/reporter.py
```diff
@@ -0,0 +1,152 @@
...
+        _log.info("wrote results.json with %d entries to %s", len(results), path)
+        return path
+
+    def write_rows(self, run_dir: str | os.PathLike[str], rows: list[dict[str, Any]]) -> Path:
+        """Write flattened ingest rows to ``<run_dir>/rows.json`` and return the path.
+
+        ``rows`` is the dashboard-facing ``ResultRow`` contract (one entry per
+        ``(setup × task × run × iteration)``), already converted to dicts via
```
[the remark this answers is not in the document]
janetkuo: The repo's [tool.ruff.lint] select set is:
```
select = ["E", "F", "I", "UP", "B", "SIM"]
```

RUF rules aren't enabled, so RUF002 will not fail CI

PR #35, opened 2026-07-21
feat(verification): add pod-healthy and scaling-complete verifiers

itssimrank:
Adds two concrete leaf verifiers registered against the verification registry. PodHealthyVerifier blocks on kubectl wait for matched pods to reach Ready, falling back to checking pod phase directly if the wait fails or times out. ScalingCompleteVerifier polls a deployment ready replica count with exponential backoff until it converges into a [min, max] range, covering both scale-up and scale-down/optimization targets.

## Summary by CodeRabbit

- **New Features**
  - Added pod health verification that confirms matching pods become ready and falls back to checking their running status when readiness checks fail.
  - Added scaling completion verification for deployments, including minimum and optional maximum replica requirements.
  - Verification results now provide clear success or failure reasons and relevant resource details.

- **Tests**
  - Added coverage for pod readiness, fallback behavior, missing status data, replica limits, and deployment retrieval failures.

### DIFF
--- a/devops_bench/verification/verifiers/__init__.py
+++ b/devops_bench/verification/verifiers/__init__.py
@@ -1,0 +1,20 @@
+# Copyright 2026 The Kubernetes Authors.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""Concrete single-condition verifiers."""
+
+from devops_bench.verification.verifiers.pod_healthy import PodHealthyVerifier
+from devops_bench.verification.verifiers.scaling_complete import ScalingCompleteVerifier
+
+__all__ = ["PodHealthyVerifier", "ScalingCompleteVerifier"]
--- a/devops_bench/verification/verifiers/pod_healthy.py
+++ b/devops_bench/verification/verifiers/pod_healthy.py
@@ -1,0 +1,126 @@
+# Copyright 2026 The Kubernetes Authors.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""Verifier that waits for selected pods to become Ready/Running."""
+
+from __future__ import annotations
+
+import time
+from typing import Any, Literal
+
+from devops_bench.core import SubprocessError, get_logger
+from devops_bench.k8s import get_resource, wait
+from devops_bench.verification.base import VERIFIERS, BaseVerifier, VerificationResult
+
+__all__ = ["PodHealthyVerifier"]
+
+_log = get_logger("verification.pod_healthy")
+
+
+@VERIFIERS.register("pod_healthy")
+class PodHealthyVerifier(BaseVerifier):
+    """Verify that pods matched by a selector are Ready (Running on fallback).
+
+    The primary path blocks on ``kubectl wait --for=condition=Ready``. If that
+    fails or times out, it falls back to inspecting pod phases and succeeds when
+    every matched pod is ``Running``.
+
+    Attributes:
+        type: Discriminator literal, always ``"pod_healthy"``.
+        selector: Label selector (``-l``) identifying the pods.
+        namespace: Optional namespace; defaults to the active one.
+    """
+
+    type: Literal["pod_healthy"] = "pod_healthy"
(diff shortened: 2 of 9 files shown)


devops_bench/verification/verifiers/pod_healthy.py
```diff
@@ -0,0 +1,126 @@
...
+                    raw=raw,
+                )
+
+            stderr = (exc.stderr or "").strip()
+            return VerificationResult(
+                success=False,
+                elapsed_time=elapsed,
+                reason=f"kubectl wait failed or timed out: {stderr}",
+                name=self.name,
+                raw=raw,
```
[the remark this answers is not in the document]
itssimrank: Updated.

tests/unit/verification/test_pod_healthy.py
```diff
@@ -0,0 +1,119 @@
...
+from __future__ import annotations
+
+from types import SimpleNamespace
+from unittest.mock import patch
+
+from devops_bench.core import SubprocessError
+from devops_bench.verification.verifiers import PodHealthyVerifier
+
+
+def test_kubectl_wait_success_returns_raw_output():
```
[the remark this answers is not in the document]
itssimrank: Updated.

devops_bench/verification/verifiers/scaling_complete.py
```diff
@@ -0,0 +1,106 @@
...
+        max_replicas: Optional maximum ready replicas allowed for success; when
+            ``None`` only the lower bound is enforced.
+        namespace: Optional namespace; defaults to the active one.
+    """
+
+    type: Literal["scaling_complete"] = "scaling_complete"
+    deployment: str
+    min_replicas: int = 1
+    max_replicas: int | None = None
+    namespace: str | None = None
```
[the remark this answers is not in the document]
itssimrank: Added a model validator for it.

devops_bench/verification/verifiers/__init__.py
```diff
@@ -0,0 +1,20 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""Concrete single-condition verifiers."""
+
+from devops_bench.verification.verifiers.pod_healthy import PodHealthyVerifier
+from devops_bench.verification.verifiers.scaling_complete import ScalingCompleteVerifier
```
jessie1111101, line 18: These register only as an import side effect, but nothing in the runtime path imports the verifiers package (verification/spec.py and __init__.py don't, and there are no devops_bench.verifiers entry points), so  VerificationSpec({"type":"pod_healthy",...}) raises unknown-verifier at runtime; the yests pass only because they import the classes directly. Add from devops_bench.verification import verifiers  # noqa: F401 to spec.py (like chaos/spec.py does), or declare entry points.

devops_bench/verification/verifiers/pod_healthy.py
```diff
@@ -0,0 +1,126 @@
...
+
+    def _check_pods_status(self, raw: dict[str, Any]) -> bool:
+        """Return True when at least one pod matched and all are ``Running``.
+
+        A pod whose ``status`` is explicitly ``null`` is treated as not Running
+        rather than crashing the check.
+        """
+        items = raw.get("items", [])
+        return len(items) > 0 and all(
+            (p.get("status") or {}).get("phase") == "Running" for p in items
```
eugeneng04: I think there is a case if the pod is in a `CrashLoopBackOff` state, it will show up as running, even though its repeateadly starting, failing and restarting over and over again. Where because we get some status, we're saying that its healthy.

Maybe `status.conditions[type==Ready].status == "True"` would be a more rigorous check, and fall back to check running if that information isn't present

devops_bench/verification/verifiers/pod_healthy.py
```diff
@@ -0,0 +1,126 @@
...
+                elapsed_time=elapsed,
+                reason=f"kubectl wait failed or timed out: {stderr}",
+                name=self.name,
+                raw=raw,
+            )
+
+    def _get_pods_details(self) -> dict[str, Any]:
+        """Fetch matched pods as JSON, returning an error dict on failure."""
+        try:
+            return get_resource(
```
eugeneng04, line 107: `get_resource` isn't given a timeout so if the server accepts connection and doesn't respond, it will hang

devops_bench/verification/verifiers/scaling_complete.py
```diff
@@ -0,0 +1,126 @@
...
+
+    def _check_scaling(self) -> tuple[bool, str, dict[str, Any] | None]:
+        """Read the deployment once and compare ready replicas to the target.
+
+        Returns:
+            A ``(success, reason, raw)`` triple. ``raw`` carries the raw
+            deployment document once one could be read, else ``None``.
+        """
+        try:
+            dep_data = get_resource(
```
eugeneng04, line 100: same thing here as above, I think we should add a timeout

devops_bench/verification/verifiers/pod_healthy.py
itssimrank: added.

devops_bench/verification/verifiers/scaling_complete.py
itssimrank: updated with timeout param

devops_bench/verification/verifiers/pod_healthy.py
itssimrank: Good callout. Updated the check to check the status condition.

devops_bench/verification/verifiers/__init__.py
itssimrank: Imported the verifiers.

janetkuo (approved):
/lgtm

PR #37, opened 2026-07-22
Add the evaluation harness (base, scenario, and default harness)

jessie1111101:
## What this adds

The evaluation harness:

- `evalharness/base.py` — the `Harness` abstract base.
- `evalharness/scenario.py` — the `ScenarioManager` (runs chaos + verification against a cluster, with port-forward helpers).
- `evalharness/artifacts.py` — workspace snapshot + generated-file collection.
- `evalharness/default.py` — `DefaultEvalHarness`: the task loop (provision → run agent → optional chaos/verify → score → teardown).
- `evalharness/__init__.py` — the package's public surface.

## Testing

Unit tests cover the base, scenario manager, and default harness (registry resolution, single-env reads, package-import laziness). `uv sync --frozen`, `ruff check`, `ruff format --check`, and `pytest` all pass. End-to-end and spec-parsing coverage that relies on concrete verifiers and task fixtures will follow as those land.

/kind feature

```release-note
NONE
```

## Summary by CodeRabbit

* **New Features**
  * Added an end-to-end evaluation harness with agent execution, scoring, and JSON report outputs.
  * Introduced a scenario orchestrator for chaos injection and optional verification with aggregated chaos/performance reporting.
  * Added workspace artifact collection for newly generated outputs.
  * Added public package exports for harness and scenario tooling.
* **Bug Fixes**
  * Improved robustness so injection/verification/setup failures are recorded per task without aborting the full run.
  * Enhanced load-balancer URL and port-forward behavior, including graceful fallbacks and clearer unresolved verification reporting.
* **Tests**
  * Added unit coverage for scenario wiring, report handling, package import cleanliness, registry resolution, and single-read env conventions.

### DIFF
--- a/devops_bench/evalharness/artifacts.py
+++ b/devops_bench/evalharness/artifacts.py
@@ -1,0 +1,100 @@
+# Copyright 2026 The Kubernetes Authors.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""Capture files an agent generates by diffing a directory before and after."""
+
+from __future__ import annotations
+
+import os
+import shutil
+from pathlib import Path
+
+from devops_bench.core import get_logger
+
+__all__ = ["snapshot_dir", "collect_generated_files"]
+
+_log = get_logger("evalharness.artifacts")
+
+
+def snapshot_dir(path: str | os.PathLike[str] = ".") -> set[str]:
+    """Snapshot the immediate entries of a directory.
+
+    Args:
+        path: Directory to list; defaults to the current working directory.
+
+    Returns:
+        The set of entry names directly under ``path``. An empty set is
+        returned when ``path`` does not exist, so the diff stays well-defined
+        when the workspace is created lazily by the agent.
+    """
+    target = os.fspath(path)
+    if not os.path.isdir(target):
+        return set()
+    return set(os.listdir(target))
+
+
+def collect_generated_files(
+    before: set[str],
+    run_dir: str | os.PathLike[str],
+    *,
+    source_dir: str | os.PathLike[str] = ".",
+) -> list[str]:
+    """Copy entries created since ``before`` into the run's artifact directory.
+
+    New files and directories (those present now but absent from ``before``) are
+    copied into ``<run_dir>/generated_files/``. The destination directory is
+    created only when there is at least one new entry to copy.
+
+    Args:
+        before: Entry names captured by :func:`snapshot_dir` prior to the run.
+        run_dir: The run output directory; artifacts land under its
+            ``generated_files`` subdirectory.
+        source_dir: Directory the agent wrote into; defaults to the current
+            working directory. The harness threads its harness-owned, per-run
+            :attr:`~devops_bench.core.RunContext.workspace_path` here — the
+            same directory the CLI agent wrapper executes in — so the
+            artifact diff is bound to the per-task workspace, not the
+            harness process's cwd.
+
+    Returns:
+        The names of the entries that were copied.
+    """
+    after = snapshot_dir(source_dir)
+    new_entries = after - before
+    if not new_entries:
+        return []
+
+    gen_files_dir = Path(run_dir) / "generated_files"
+    gen_files_dir.mkdir(parents=True, exist_ok=True)
+
+    copied: list[str] = []
+    src_root = os.fspath(source_dir)
+    for name in new_entries:
+        src = os.path.join(src_root, name)
+        dst = os.fspath(gen_files_dir / name)
+        # An agent could plant a symlink pointing outside the workspace; following
+        # it would copy host files/credentials into the run artifacts. Skip links
+        # at the top level and preserve (don't follow) any nested ones.
+        if os.path.islink(src):
+            _log.warning("skipping symlinked artifact %s", src)
+            continue
+        if os.path.isdir(src):
+            shutil.copytree(src, dst, dirs_exist_ok=True, symlinks=True)
+            copied.append(name)
+        elif os.path.isfile(src):
+            shutil.copy(src, dst)
+            copied.append(name)
+
+    _log.info("collected %d generated artifact(s) into %s", len(copied), gen_files_dir)
+    return copied
--- a/devops_bench/evalharness/default.py
+++ b/devops_bench/evalharness/default.py
@@ -1,0 +1,1090 @@
+# Copyright 2026 The Kubernetes Authors.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""DefaultEvalHarness: wires agents, chaos, verification, and metrics into one pipeline."""
+
+from __future__ import annotations
+
+import datetime
+import importlib
+import json
+import shutil
+import tempfile
+import threading
+from dataclasses import replace
+from pathlib import Path
+from typing import Any
+
+from devops_bench.agents import AGENTS, AgentConfig, AgentResult
+from devops_bench.agents.capabilities import (
+    AgentRules,
+    AllCapabilities,
+    McpBinding,
+    SkillBinding,
+)
+from devops_bench.chaos import ChaosSpec
+from devops_bench.core import (
+    ConfigError,
+    MissingDependencyError,
+    NotRegisteredError,
+    RunContext,
+    get_bool,
+    get_env,
+    get_logger,
+)
+from devops_bench.deployers.factory import get_deployer
+from devops_bench.evalharness.artifacts import collect_generated_files, snapshot_dir
+from devops_bench.evalharness.base import Harness
+from devops_bench.evalharness.reporter import ResultReporter
+from devops_bench.evalharness.scenario import (
+    VERIFICATION_TIMEOUT_SEC,
+    ScenarioManager,
+    pick_free_port,
+)
+from devops_bench.tasks import Task
+from devops_bench.verification import VerificationSpec
+
+__all__ = ["DefaultEvalHarness"]
+
+_log = get_logger("evalharness.default")
+
+# Builtin agent modules imported at call time so their ``@AGENTS.register``
+# decorators run. External packages add agents by registering with the same
+# registry, with no edit here.
+_BUILTIN_AGENT_MODULES: tuple[str, ...] = (
+    "devops_bench.agents.cli.gemini_cli",
+    "devops_bench.agents.cli.openclaw",
+    "devops_bench.agents.cli.antigravity",
+    "devops_bench.agents.api.agent",
+)
+
+# Aliases normalized to canonical agent keys before registry lookup.
+_AGENT_TYPE_ALIASES: dict[str, str] = {
+    "gemini-cli": "gemini",
+}
+
+# Default agent type when neither --agent-type nor BENCH_AGENT_TYPE is set.
+_DEFAULT_AGENT_TYPE = "gemini-cli"
+
+# Default target deployment + namespace used both for placeholder
+# substitution in the agent prompt and as the chaos port-forward target, so the
+# operator agent and the chaos injector address the same workload when env is
+# unset.
+_DEFAULT_TARGET_DEPLOYMENT = "hypercomputer-d1-frontend"
+_DEFAULT_NAMESPACE = "default"
+
+# How long to wait for the chaos agent to establish its load spike before
+# starting the operator agent.
+_CHAOS_ACTIVE_WAIT_SEC = 45
+
+# Budget for draining the scenario thread. Kept above the verification budget
+# so a slow-but-completing verification is not cut off, which would otherwise
+# yield partial reports and race teardown.
+_SCENARIO_JOIN_SEC = VERIFICATION_TIMEOUT_SEC + 60
+
+
+def _ensure_builtin_agents_registered() -> None:
+    """Import the builtin agent modules so their registrations fire.
+
+    The registry is the only source of truth — this function exists so the
+    harness can resolve canonical keys at call time without naming any module
+    path in ``AGENTS.get``. Re-imports are no-ops thanks to ``sys.modules``.
+
+    Catches **only** missing-dependency / import errors (an agent module may
+    pull an optional SDK like ``anthropic`` that is absent on the host) — a
+    real bug in an agent module (``SyntaxError``, an ``AttributeError`` at
+    module top) re-raises so it cannot hide behind a silent ``debug`` log.
+    """
+    for module in _BUILTIN_AGENT_MODULES:
+        try:
+            importlib.import_module(module)
+        except (ImportError, MissingDependencyError) as exc:
+            # Optional SDK absent on this host. ``AGENTS.get`` will still
+            # raise a clear ``NotRegisteredError`` later if the user selects
+            # an agent whose module did not load.
+            _log.debug("optional agent module %s not importable: %s", module, exc)
+
+
+class DefaultEvalHarness(Harness):
+    """Standard harness wiring every component into one pipeline.
+
+    Each task flows through provisioning, optional background chaos, agent
+    execution, artifact collection, teardown, and batch scoring. Every layer
+    is consumed through its typed contract: ``Task`` in, ``AgentResult`` from
+    the agent, ``ChaosResult`` / ``VerificationResult`` from the scenario,
+    ``MetricScore`` from each metric. The harness routes those typed values
+    through ``to_dict()`` / ``to_entry()`` / ``model_dump()`` so the on-disk
+    ``results.json`` schema stays byte-stable.
+
+    Args:
+        project_id: Default GCP project ID for provisioning and placeholders.
+        cluster_name: Default cluster name for provisioning and placeholders.
+        judge_model: A ``DeepEvalBaseLLM`` judge used for scoring; when ``None``
+            one is built from ``JUDGE_PROVIDER`` / ``JUDGE_MODEL`` on first use.
+        results_root: Directory under which timestamped run dirs are created.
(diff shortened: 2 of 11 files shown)


devops_bench/evalharness/artifacts.py
```diff
@@ -0,0 +1,94 @@
...
+    copied: list[str] = []
+    src_root = os.fspath(source_dir)
+    for name in new_entries:
+        src = os.path.join(src_root, name)
+        dst = os.fspath(gen_files_dir / name)
+        if os.path.isdir(src):
+            shutil.copytree(src, dst, dirs_exist_ok=True)
+            copied.append(name)
+        elif os.path.isfile(src):
+            shutil.copy(src, dst)
```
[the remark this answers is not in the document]
jessie1111101: Done — skip top-level symlinked entries (warn + continue) before the isdir/isfile checks, and pass `symlinks=True` to `copytree` so nested links aren't followed either.

tests/unit/evalharness/test_scenario.py
```diff
@@ -0,0 +1,570 @@
...
+            target_deployment="dep", namespace="ns", skip_port_forward=True
+        ).run_chaos_and_verification(spec, _build_ctx())
+
+    assert _ENV_LOCAL_PORT not in captured["env"]
+
+
+def test_pick_free_port_returns_distinct_usable_ports() -> None:
+    port = pick_free_port()
+    assert isinstance(port, int)
+    assert 1 <= port <= 65535
```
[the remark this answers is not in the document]
jessie1111101: Renamed to `test_pick_free_port_returns_a_usable_port` — `pick_free_port` binds port 0 and releases it, so distinctness isn't guaranteed and asserting it would be flaky; noted that in a comment.

tests/unit/evalharness/test_scenario.py
```diff
@@ -0,0 +1,570 @@
...
+    tunnel, but the guard pins the invariant explicitly: it patches the
+    ``subprocess.Popen`` the port-forward helper would use so a future refactor
+    that forgets the flag fails loudly instead of attempting a real
+    port-forward.
+    """
+
+    def _boom(*args, **kwargs):  # pragma: no cover - exercised only on regression
+        raise RuntimeError("test attempted to spawn a real kubectl process")
+
+    monkeypatch.setattr("devops_bench.k8s.kubectl.subprocess.Popen", _boom)
```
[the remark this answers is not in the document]
jessie1111101: Fixed — dropped the 'every test' claim and reworded to note that tests not driving the port-forward path set `skip_port_forward=True`, with this guard backstopping the rest.

devops_bench/evalharness/scenario.py
```diff
@@ -0,0 +1,436 @@
...
+            self.result_holder["chaos_report"]["verification"] = verification_result.model_dump()
+            self.result_holder["perf_report"] = self._perf_from_verification(verification_result)
+        except Exception as exc:  # noqa: BLE001 - surface failure into the report
+            _log.error("verification failed with exception: %s", exc)
+            self.result_holder["chaos_report"]["verification"] = {
+                "success": False,
+                "reason": f"Verification exception: {exc}",
+            }
+
+    def _inject_chaos(self, spec: ChaosSpec, ctx: RunContext):
```
itssimrank: nit: missing its return type

tests/unit/evalharness/test_single_env_read.py
```diff
@@ -0,0 +1,177 @@
...
+    config = harness.build_agent_config()
+    assert config.capabilities.mcp_servers == ()
+    assert config.capabilities.tools_enabled is False
+    assert harness.use_mcp is False
+
+    # Metrics side: the score path threads the same boolean as
+    # ``use_mcp=False`` into ``evaluate_metrics_batch``.
+    captured: dict = {}
+
+    def fake_batch(results, judge, *, use_mcp):
```
itssimrank: nit: missing type annotations

devops_bench/evalharness/scenario.py
```diff
@@ -0,0 +1,436 @@
...
+        self.target_deployment = target_deployment
+        self.namespace = namespace
+        self.verification_mapping: dict[str, Any] = dict(verification_mapping or {})
+        self.skip_port_forward = skip_port_forward
+        # Per-run local port for the fault's port-forward; None keeps the
+        # fault's default. Parallel runs pass a free port to avoid contention.
+        self.local_port = local_port
+        self.chaos_active_event = threading.Event()
+        self.verifier_agent = VerifierAgent()
+        self.result_holder: dict[str, dict[str, Any]] = {
```
itssimrank, line 134: Should we wrap this in a light threading.Lock to prevent thread-safety issues?

devops_bench/evalharness/default.py
```diff
@@ -0,0 +1,1042 @@
...
+            node = entry.get("spec") if "spec" in entry else entry
+            try:
+                mapping[name] = VerificationSpec(node)
+            except Exception as exc:  # noqa: BLE001 - surface every failure
+                _log.warning(
+                    "verification entry %r failed to validate; skipping: %s",
+                    name,
+                    exc,
+                )
+                errors.append({"name": str(name), "reason": str(exc)})
```
itssimrank, line 520: This could generate multi-line text strings. Should we consider serializing structured error summaries?

devops_bench/evalharness/default.py
```diff
@@ -0,0 +1,1042 @@
...
+                _log.info("waiting for chaos agent to establish the cluster load spike...")
+                scenario_manager.chaos_active_event.wait(timeout=_CHAOS_ACTIVE_WAIT_SEC)
+                _log.info("cluster load spike active; proceeding with operator agent...")
+
+            _log.info("executing agent for prompt: %s", prompt)
+            before_files = snapshot_dir(workspace_path)
+            agent_res = self.execute_agent(prompt, context)
+            # NOTE/TODO: This collects ALL frontmatter from bootstrapping, not just generated files.
+            # Consider a more targeted filter in a future iteration.
+            collect_generated_files(before_files, run_dir, source_dir=workspace_path)
```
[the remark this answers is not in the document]
jessie1111101: Done. Wrapped `collect_generated_files` in a try/except so a collection failure logs and continues instead of turning a completed agent run into a failed, unscored record. This matches the best effort isolation the rest of the file already uses.

devops_bench/evalharness/default.py
```diff
@@ -0,0 +1,1042 @@
...
+        result: dict[str, Any] | None = None
+        workspace_path: Path | None = None
+        verification_parse_errors: list[dict[str, str]] = []
+        # Track the substituted prompt / expectation as they are computed so a
+        # failed record can carry the same resolved strings a success record
+        # would, falling back to the raw task fields before substitution.
+        prompt: str | None = None
+        expected_output: str | None = None
+
+        try:
```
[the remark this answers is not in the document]
jessie1111101: Done. After `stop()` the finally now joins the scenario thread with a bounded timeout before teardown, so a failure path no longer races a still running background scenario. The success path already drains and joins via `_drain_scenario`; this covers the exception path too.

devops_bench/evalharness/default.py
jessie1111101: Good point. The entry is already a structured dict (name plus reason), but reason is `str(exc)`, which for a pydantic error can span several lines. Happy to normalize that (collapse whitespace, or keep the first line) if you prefer. Fully structuring pydantic error detail feels like a separate follow up, so I left it as is for now.

devops_bench/evalharness/scenario.py
jessie1111101: `result_holder` is written only by the scenario thread and read by the main thread only after the thread is joined (via `_drain_scenario` on success, and now the finally in `default.py` on the exception path). `Thread.join` gives the happens before barrier, so a lock would be redundant here. If you spot a path that reads it without joining first, I will add one.

tests/unit/evalharness/test_single_env_read.py
jessie1111101: Done. Annotated the `fake_batch` stub (its params plus a None return).

devops_bench/evalharness/scenario.py
jessie1111101: Done. Added the None return annotation on `_inject_chaos`.

devops_bench/evalharness/default.py
```diff
@@ -0,0 +1,1054 @@
...
+                _SCENARIO_JOIN_SEC,
+            )
+            # The daemon thread is still running and may be mid-write, so deep-
+            # copy the live report before stamping it: a shallow ``dict()`` would
+            # share nested objects and could capture a torn structure. This
+            # preserves any partial fields populated before the cutoff
+            # (injected_fault / name / output) so the operator sees how far it got.
+            chaos_report = copy.deepcopy(chaos_report)
+            chaos_report["status"] = "timed_out"
+        return chaos_report, perf_report
```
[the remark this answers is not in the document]
jessie1111101: Good catch. `get_reports()` now takes the report lock and returns a deep copy, and every `result_holder` write in the scenario thread takes the same lock, so the timeout snapshot in `_drain_scenario` can no longer race a mid write. Also dropped the now redundant `deepcopy` in `default.py` since the snapshot is already private.

devops_bench/evalharness/scenario.py
jessie1111101: Correction to my earlier reply: you were right. The normal path joins before reading, but the timeout branch in `_drain_scenario` reads while the thread may still be running (the join can time out), so there is a real race there. Added a `threading.Lock` in ScenarioManager guarding the `result_holder` writes, and `get_reports()` now returns a deep copy under that lock. Thanks for flagging it.

devops_bench/evalharness/default.py
```diff
@@ -0,0 +1,1052 @@
...
+        if not raw:
+            return []
+        resolved = self._resolve_spec_placeholders(raw, cluster_name, target_deployment, namespace)
+        # A placeholder-substituted JSON string round-trips through
+        # ``json.loads`` to a list/dict the discriminated union can validate.
+        if isinstance(resolved, str):
+            try:
+                resolved = json.loads(resolved)
+            except json.JSONDecodeError as exc:
+                _log.warning("could not parse chaos_spec JSON string: %s", exc)
```
pradeepvrd: We should likely raise an exception here. This can just skip a chaos trigger completely and proceed further and silently make the result invalid.

devops_bench/evalharness/default.py
```diff
@@ -0,0 +1,1052 @@
...
+                self._teardown(deployer, infra_config, task.name)
+            if workspace_path is not None:
+                shutil.rmtree(workspace_path, ignore_errors=True)
+
+        return result
+
+    #: Top-level keys present on **every** record (success and failed alike).
+    #: Exposed so downstream parsers / tests can pin the symmetric schema
+    #: without reproducing the literal in every spot.
+    _RECORD_KEYS: frozenset[str] = frozenset(
```
pradeepvrd, line 824: Is this actually used anywhere?

devops_bench/evalharness/default.py
```diff
@@ -0,0 +1,1052 @@
...
+            "perf_report": {},
+            "documentation": [doc.model_dump() for doc in task.documentation],
+            "capabilities_granted": {
+                "use_mcp": self.use_mcp,
+                "skills": list(self._granted_skill_paths),
+            },
+            "verification_parse_errors": [],
+            # Generation-only (``deployer: noop``) tasks have no cluster, so the
+            # OutcomeValidity judge must not penalize them for "not applying".
+            "generation_only": (task.infrastructure or {}).get("deployer") == "noop",
```
pradeepvrd: This is missing enforcing the no_infra condition. Should be

```
self.no_infra or (task.infrastructure or {}).get("deployer") == "noop"
```

devops_bench/evalharness/default.py
```diff
@@ -0,0 +1,1052 @@
...
+                :class:`~devops_bench.deployers.NoOpDeployer`.
+
+        Returns:
+            A ``(scenario_manager, thread)`` pair, or ``None`` when no chaos
+            specs were provided.
+        """
+        if not chaos_specs:
+            return None
+
+        spec = chaos_specs[0]
```
pradeepvrd, line 566: Note that the spec has now changed to make this is a list but we only trigger the first spec. We should either make this field in the task file a singleton or handle the different conditions. If this is an array for future extension, a clear warning message would be good.

devops_bench/evalharness/default.py
```diff
@@ -0,0 +1,1052 @@
...
+                chaos_specs,
+                verification_mapping,
+                replace(context, env=dict(context.env)),
+                target_deployment=target_dep,
+                namespace=ns,
+            )
+            if scenario is not None:
+                scenario_manager, scenario_thread = scenario
+                _log.info("waiting for chaos agent to establish the cluster load spike...")
+                scenario_manager.chaos_active_event.wait(timeout=_CHAOS_ACTIVE_WAIT_SEC)
```
pradeepvrd: What if the wait condition fails?

devops_bench/evalharness/default.py
```diff
@@ -0,0 +1,1052 @@
...
+            if not name:
+                msg = f"verification_spec entry [{index}] missing required ``name`` key"
+                _log.warning(msg)
+                errors.append({"name": f"<index {index}>", "reason": msg})
+                continue
+            # Canonical shape ``{name, spec: <typed-node>}``; an entry without
+            # a ``spec`` key is itself treated as the typed node.
+            node = entry.get("spec") if "spec" in entry else entry
+            try:
+                mapping[name] = VerificationSpec(node)
```
pradeepvrd, line 513: We would always pick the latest spec with a name if there is name collision and silently drop the earlier ones.

We should return a duplicate name error.

devops_bench/evalharness/scenario.py
```diff
@@ -0,0 +1,456 @@
...
+                )
+        except Exception as exc:  # noqa: BLE001 - surface failure into the report
+            _log.error("verification failed with exception: %s", exc)
+            with self._report_lock:
+                self.result_holder["chaos_report"]["verification"] = {
+                    "success": False,
+                    "reason": f"Verification exception: {exc}",
+                }
+
+    def _inject_chaos(self, spec: ChaosSpec, ctx: RunContext) -> None:
```
pradeepvrd: Incorrect ReturnType. Should be `ChaosResult`

devops_bench/evalharness/default.py
jessie1111101: Agreed. Changed it to raise `ConfigError` instead of returning an empty list, so a declared but unparseable chaos_spec fails the task loudly (surfaced as a failed record) rather than silently running with no chaos and scoring an invalid result. Added a test that a malformed chaos_spec raises.

devops_bench/evalharness/scenario.py
```diff
@@ -0,0 +1,456 @@
...
+                # action's load URL to point at it directly. Fall back to the
+                # port-forward path if resolution fails or times out — that way a
+                # delayed LB still produces a load attempt instead of an aborted
+                # run.
+                lb_ip = self._resolve_lb_ip(self.target_deployment, self.namespace)
+
+            if lb_ip is not None and hasattr(spec.action, "target"):
+                target = spec.action.target
+                if hasattr(target, "service_url"):
+                    lb_url = f"http://{lb_ip}:{_LOCAL_PORT}"
```
pradeepvrd, line 279: The local port may not be the same as the Service port. We should parsing the port from the service definition.

devops_bench/evalharness/scenario.py
```diff
@@ -0,0 +1,456 @@
...
+
+        # Thread the port-forward target onto the context. ``ctx.env`` values
+        # are strings; flags are written only when truthy so the fault's
+        # ``bool(env.get(...))`` reads cleanly.
+        ctx.env[_ENV_TARGET_DEPLOYMENT] = self.target_deployment
+        ctx.env[_ENV_TARGET_NAMESPACE] = self.namespace
+        if self.local_port is not None:
+            ctx.env[_ENV_LOCAL_PORT] = str(self.local_port)
+
+        if self.skip_port_forward:
```
pradeepvrd, line 254: This should belong with `GenerateLoadFault` actually. The _inject_chaos function should be independent of trigger specific steps.

The fact that we are depending on private variables from that modules is a clear sign we are leaking semantics.

devops_bench/evalharness/default.py
jessie1111101: It's being used in tests only. Maybe we can remove it in a follow up PR?

devops_bench/evalharness/default.py
jessie1111101: good catch added back. generation_only is now self.no_infra or (task.infrastructure or {}).get("deployer") == "noop", so a whole run under BENCH_NO_INFRA is flagged generation_only just like a per task noop deployer. This mirrors get_deployer's own gate, and the golden tests cover generation_only being present in the key set.

devops_bench/evalharness/default.py
jessie1111101: Good catch. The scheduler still runs only the first spec, but no longer silently: if chaos_spec declares more than one entry we log a warning naming how many are ignored, so a task authored with several disruptions is not quietly under run. Making the field a true singleton, or scheduling every entry, is a larger behavior change I would rather do as a follow up PR

devops_bench/evalharness/default.py
jessie1111101: Handled. chaos_active_event.wait now captures its return value. The event is also set when injection fails (to unblock the operator), so a False return means chaos never signalled active within _CHAOS_ACTIVE_WAIT_SEC. In that case we log a warning that the run may not reflect the intended disruption and proceed; the drained chaos_report still carries the failure detail for the judge. We keep running rather than aborting so a slow or failed spike degrades to a best effort run instead of losing the record entirely.

devops_bench/evalharness/default.py
jessie1111101: SG, added a duplicate name guard in _build_verification_mapping: if a name is already in the mapping we skip the entry and record a parse error (name plus reason) instead of letting the later spec silently overwrite the earlier one. The collision now surfaces on the record as a verification_parse_error rather than being dropped.

devops_bench/evalharness/scenario.py
jessie1111101: good catch, the last edit incorrectly renamed it. fixed now

devops_bench/evalharness/scenario.py
jessie1111101: Agreed, this is a real coupling issue: _inject_chaos reaches into GenerateLoadFault internals (the _ENV_ flags and the LB resolution) instead of staying trigger agnostic. Maybe will address this in a follow up PR that moves the LB resolution and env wiring into the fault and gives the action a small public interface?

devops_bench/evalharness/scenario.py
jessie1111101: Good point. The LB URL hardcodes _LOCAL_PORT (8080) and assumes the Service port matches; it should read the port from the Service definition (spec.ports) instead. This is the same GenerateLoadFault coupling as above and is likewise carried over verbatim from the ported source, so I would like to fold it into that same follow up PR. wdyt

pradeepvrd (approved):
LGTM. thanks for addressing all comments

janetkuo (approved):
/lgtm

PR #40, opened 2026-07-23
feat(agents): add the API agent with MCP and skills support

richackard:
Adds an API-driven agent harness on the shared tool-call loop: an async stdio MCP client, local SKILL.md skills advertised as synthetic tools, trajectory folding with FIFO pairing for id-less provider tool calls, provider token extraction, and a whole-run wall-clock timeout.

Skill discovery lives in a shared iter_skills walk (expanduser, sorted order, escaping and duplicate names warned and skipped) consumed by both the CLI skill materializer and the new agent, so discovery semantics cannot diverge per harness. Adds the mcp dependency.

## Summary by CodeRabbit

* **New Features**
  * Added an API agent capable of using local skills and connected tool servers.
  * Added automatic skill discovery from configured directories, including descriptions and content.
  * Added support for tool-call tracking and token usage reporting across providers.
* **Bug Fixes**
  * Tool failures are now reported without stopping the entire run.
  * Improved handling of missing tools, unmatched results, timeouts, and invalid configurations.
  * Heavy runtime integrations load only when needed, improving startup behavior.

### DIFF
--- a/devops_bench/agents/api/__init__.py
+++ b/devops_bench/agents/api/__init__.py
@@ -1,0 +1,29 @@
+# Copyright 2026 The Kubernetes Authors.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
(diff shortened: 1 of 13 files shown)


richackard:
@coderabbitai review

janetkuo (approved):
/lgtm

PR #41, opened 2026-07-23
Add the CLI entrypoint and benchmark runner

pradeepvrd:
Adds the command-line entrypoint for running the benchmark:

- `devops_bench/run.py` — `run_benchmark()` loads tasks, builds the eval harness, and reports results. Heavy imports stay function-local so `import devops_bench.run` stays light.
- `devops_bench/cli.py` + `devops_bench/__main__.py` — argument parsing and `python -m devops_bench`.
- A `devops-bench` console script in `pyproject.toml`, and the package-root re-export of `run_benchmark`.
- Unit tests under `tests/unit/run/`.

Depends on #37 (merged): `run_benchmark` drives `DefaultEvalHarness`.

## Summary by CodeRabbit

- **New Features**
  - Added a `devops-bench` command-line interface and `python -m devops_bench` entrypoint for running benchmarks with selection, limits, infra/teardown toggles, parallelism, and run isolation.
  - Exposed benchmark configuration/results/execution APIs at the package level and clarified public exports.
  - Improved run environment signaling (including `BENCH_PARALLEL`) and ensured environment changes are reverted after runs.
  - Added run summary output with distinct exit codes for config errors vs task failures.
- **Tests**
  - Expanded unit coverage for CLI precedence/error handling, environment resolution, SDK import laziness, run execution/path reporting, and environment apply/restore behavior.

### DIFF
--- a/devops_bench/cli.py
+++ b/devops_bench/cli.py
@@ -1,0 +1,171 @@
+# Copyright 2026 The Kubernetes Authors.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""Thin argparse CLI wrapping :func:`devops_bench.run.run_benchmark`.
+
+The library entrypoint owns all real work; this module only parses flags into a
+:class:`~devops_bench.run.BenchmarkConfig` and maps the outcome to an exit code.
+``run`` imports stay function-local so ``import devops_bench.cli`` stays light.
+"""
+
+from __future__ import annotations
+
+import argparse
+import sys
+from dataclasses import replace
+from typing import TYPE_CHECKING
+
+from devops_bench.core import ConfigError
+
+if TYPE_CHECKING:  # pragma: no cover - typing-only imports
+    from devops_bench.run import BenchmarkConfig
+
+
+def build_parser() -> argparse.ArgumentParser:
+    """Build the argument parser for the ``devops-bench`` console script.
+
+    Returns:
+        A configured :class:`argparse.ArgumentParser`.
+    """
+    parser = argparse.ArgumentParser(
+        prog="devops-bench",
+        description="Run the DevOps-bench evaluation pipeline over a task source.",
+    )
+    parser.add_argument(
+        "source",
+        help="Tasks directory or task spec file (.yaml / .yml / .json).",
+    )
+    parser.add_argument("--project", dest="project_id", default=None, help="Cloud project id.")
+    parser.add_argument("--cluster", dest="cluster_name", default=None, help="Target cluster name.")
+    parser.add_argument("--limit", type=int, default=None, help="Run only the first N tasks.")
+    parser.add_argument("--results-root", default=None, help="Root directory for run artifacts.")
+    parser.add_argument("--agent-type", default=None, help="Override BENCH_AGENT_TYPE.")
+    parser.add_argument("--judge-provider", default=None, help="Override JUDGE_PROVIDER.")
+    parser.add_argument("--judge-model", default=None, help="Override JUDGE_MODEL.")
+    # Tri-state (default None) so a flag can override the env in either
+    # direction; store_true could only ever turn these on, never back off.
+    parser.add_argument(
+        "--no-infra",
+        dest="no_infra",
+        action="store_const",
+        const=True,
+        default=None,
+        help="Skip infrastructure provisioning (no project/cluster required).",
+    )
+    parser.add_argument(
+        "--infra",
+        dest="no_infra",
+        action="store_const",
+        const=False,
+        help="Force infrastructure provisioning even if BENCH_NO_INFRA is set.",
+    )
+    parser.add_argument(
+        "--no-teardown",
+        dest="no_teardown",
+        action="store_const",
+        const=True,
+        default=None,
+        help="Skip teardown of provisioned infrastructure.",
+    )
+    parser.add_argument(
+        "--teardown",
+        dest="no_teardown",
+        action="store_const",
+        const=False,
+        help="Force teardown even if BENCH_NO_TEARDOWN is set.",
+    )
+    parser.add_argument(
+        "--parallel",
+        action="store_true",
+        help=(
+            "Isolate this run (own kubeconfig / gcloud config / tofu data dir and "
+            "a run-unique cluster name) so it can run concurrently with others."
+        ),
+    )
+    parser.add_argument(
+        "--run-id",
+        dest="run_id",
+        default=None,
+        help="Explicit run id for isolation/artifact naming (default: RUN_ID env or generated).",
+    )
+    return parser
+
+
+def args_to_config(args: argparse.Namespace) -> BenchmarkConfig:
+    """Merge parsed CLI args over an env-resolved base config.
+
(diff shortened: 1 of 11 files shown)


pradeepvrd:
/hold
Waiting on #37.

tests/unit/run/test_run_benchmark.py
```diff
@@ -0,0 +1,147 @@
...
+        self.task_count: int | None = None
+        FakeHarness.instances.append(self)
+
+    def run(self, tasks: list[object]) -> list[dict[str, object]]:
+        self.task_count = len(tasks)
+        self.reporter.new_run_dir()  # set reporter.last_run_dir
+        return list(_CANNED_RESULTS)
+
+
+def _fake_load_tasks(count: int):
```
itssimrank: nit: missing return type

devops_bench/cli.py
```diff
@@ -0,0 +1,168 @@
...
+    """
+    from devops_bench.run import run_benchmark
+
+    parser = build_parser()
+    args = parser.parse_args(argv)
+    config = args_to_config(args)
+
+    try:
+        result = run_benchmark(config)
+    except ConfigError as exc:
```
itssimrank, line 165: Should we also have an exception for FileNotFoundError to distinguish between a task failure and an input path / file-parsing error?

devops_bench/__init__.py
```diff
@@ -12,4 +12,12 @@
...
 
+"""devops-bench: a standardized benchmarking suite for DevOps agents and models."""
+
+from __future__ import annotations
+
+from devops_bench.run import BenchmarkConfig, BenchmarkResult, run_benchmark
+
 __version__ = "0.1.0"
+
+__all__ = ["__version__", "BenchmarkConfig", "BenchmarkResult", "run_benchmark"]
```
itssimrank, line 23: nit: Missing Type Annotations on Exports.

devops_bench/run.py
```diff
@@ -0,0 +1,196 @@
...
+            "PROJECT_ID/GCP_PROJECT_ID and CLUSTER_NAME/GKE_CLUSTER_NAME must be "
+            "set (or pass --no-infra / BENCH_NO_INFRA=true)"
+        )
+    project_id = config.project_id or "no-infra-project"
+    cluster_name = config.cluster_name or "no-infra-cluster"
+
+    # Establish per-run isolation BEFORE any provisioning so every gcloud /
+    # kubectl / tofu / agent subprocess inherits the run-scoped kubeconfig,
+    # gcloud config, and tofu data dir. A no-op unless ``parallel`` is set.
+    run_env = RunEnv.create(parallel=config.parallel, run_id=config.run_id)
```
jessie1111101, line 163: parallel never reaches the harness. It isolates infra state here, but DefaultEvalHarness has no parallel kwarg and reads BENCH_PARALLEL from env, which apply() does not set. So --parallel (flag only) leaves harness.parallel False and the chaos port forward falls back to the fixed 8080 port, colliding across concurrent runs on one host. Could export BENCH_PARALLEL in apply() alongside RUN_ID, or set it here before constructing the harness.

devops_bench/run.py
```diff
@@ -0,0 +1,196 @@
...
+        config.results_root, run_id=run_env.run_id if run_env.isolated else None
+    )
+    harness = DefaultEvalHarness(
+        project_id,
+        cluster_name,
+        judge_model=judge,
+        results_root=config.results_root,
+        reporter=reporter,
+        agent_type=config.agent_type,
+        no_infra=config.no_infra,
```
jessie1111101: no_infra is resolved two ways: validation on line 141 uses config.no_infra or get_bool("BENCH_NO_INFRA"), but the harness gets raw config.no_infra. Via the run_benchmark API a hand built config with BENCH_NO_INFRA=true in env skips validation yet still tells the harness to provision (same for no_teardown on 181). Passing the resolved value fixes both.

tests/unit/run/test_run_benchmark.py
pradeepvrd: Fixed

devops_bench/cli.py
pradeepvrd: The loader already normalizes missing paths/parse failures to ConfigError and calls out failure mode.

I added a test to verify this behavior.

devops_bench/__init__.py
pradeepvrd: In general, we do not add type annotations for `__all__` and I think all type checks ignore this and look at the module itself to verify type enforcements.

devops_bench/run.py
pradeepvrd: Good catch. Fixed

devops_bench/run.py
pradeepvrd: Good catch! Fixed

itssimrank (approved):

devops_bench/run.py
```diff
@@ -0,0 +1,200 @@
...
+        ``results.json`` path.
+
+    Raises:
+        ConfigError: If infrastructure is enabled but project id / cluster name
+            are missing, or if ``config.source`` does not exist.
+    """
+    # Resolve the toggles once so validation and the harness observe the same
+    # values: a hand-built config with only the env set must not skip
+    # validation here yet still tell the harness to provision (or tear down).
+    no_infra = config.no_infra or get_bool("BENCH_NO_INFRA")
```
janetkuo: Checking `get_bool("BENCH_NO_INFRA")` and `get_bool("BENCH_NO_TEARDOWN")` here again overrides the explicit configuration settings `config.no_infra` and `config.no_teardown`.

devops_bench/run.py
```diff
@@ -0,0 +1,200 @@
...
+
+
+@dataclass(frozen=True)
+class BenchmarkConfig:
+    """Resolved configuration for a single benchmark run.
+
+    Attributes:
+        source: Tasks directory or task spec file (``.yaml`` / ``.yml`` / ``.json``).
+        project_id: GCP project id; required unless infra is disabled.
+        cluster_name: GKE cluster name; required unless infra is disabled.
```
janetkuo: Is this GKE specific? It should be vendor neutral.

janetkuo: Please check other changes in the RP and fix them too

jessie1111101 (approved):

devops_bench/run.py
pradeepvrd: Good point. Fixed by making the config authoritative.

devops_bench/run.py
pradeepvrd: reworded here and swept the rest of the PR.

I did keep the GCP_PROJECT_ID variable explicit because it is needed by Vertex Auth for Anthropic/Gemini Providers.

devops_bench/run.py
```diff
@@ -0,0 +1,219 @@
...
+    run_id: str | None = None
+
+    @classmethod
+    def from_env(cls, source: str, *, env: Mapping[str, str] | None = None) -> BenchmarkConfig:
+        """Build a config from ``source`` plus environment variables.
+
+        This is the only place environment variables are folded into a config;
+        :func:`run_benchmark` treats the config it receives as authoritative.
+        ``GCP_PROJECT_ID`` is accepted as a fallback for ``PROJECT_ID`` because
+        the provider layer already reads it (GCP provisioning and Vertex AI
```
janetkuo: To maintain vendor neutrality in the core runner, we should avoid reading GCP-specific environment variables (GCP_PROJECT_ID, GKE_CLUSTER_NAME) directly in the generic framework layer.

Could we instead extend ClusterInfo in devops_bench/core/context.py, and populate them in GcpProvider in devops_bench/providers/gcp.py?

pradeepvrd: Done. dropped the GCP_PROJECT_ID fallback, with a test to ensure that provider-specific variables are ignored at this layer. 

The provider layer implements the fallback you describe: GcpProvider falls back to GCP_PROJECT_ID itself (providers/gcp.py). Please let me know if this works.

janetkuo (approved):
/lgtm

PR #42, opened 2026-07-27
Remove the unused _RECORD_KEYS constant

jessie1111101:
`DefaultEvalHarness._RECORD_KEYS` is never read at runtime. Both record builders
construct their key set from the `_empty_record` literal, so the frozenset was a
second, hand-maintained copy of the schema that could silently drift from the
records it claimed to describe.

This drops the constant and the two docstring references to it. The
symmetric-shape invariant stays pinned by the golden tests in
`test_default_harness.py`, which assert that `_build_success_record` and
`_build_failed_record` emit exactly the same expected key set.

Follow-up to the review discussion on #37.

/kind cleanup

## Summary by CodeRabbit

* **Refactor**
  * Simplified internal record schema documentation while preserving the existing output structure for successful and failed results.

* **Tests**
  * Clarified documentation for the results file schema.
  * Continued validating that successful and failed records contain matching required fields.

### DIFF
--- a/devops_bench/evalharness/default.py
+++ b/devops_bench/evalharness/default.py
@@ -818,39 +818,6 @@ class DefaultEvalHarness(Harness):
 
         return result
 
-    #: Top-level keys present on **every** record (success and failed alike).
-    #: Exposed so downstream parsers / tests can pin the symmetric schema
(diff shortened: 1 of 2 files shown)


janetkuo (approved):
/lgtm

PR #49, opened 2026-07-29
feat(deployers): allow overriding the TF stack root via BENCH_TF_ROOT

eugeneng04:
## What

Make the OpenTofu deployer's stack root overridable via `BENCH_TF_ROOT` (vendor-neutral `BENCH_*` convention, read through the core `get_env` helper so blank means unset). A relative `stack:` in `task.yaml` now resolves against `$BENCH_TF_ROOT` when set, else the checkout's `<repo_root>/tf` exactly as before. The resolved root is threaded into `_isolated_work_dir`, so stacks under an overridden root keep per-run isolation for parallel runs.

## Why

`tofu.py` resolved relative stack names against `Path(__file__).resolve().parents[2] / "tf"`. Under a pip install that lands in `site-packages`, and `tf/` is not packaged — so every relative stack fails for library consumers (first: kube-agents, which authors stacks in its own repo). The existing escape hatch, absolute `stack:` paths, skips per-run isolation, so concurrent runs sharing one stack dir contend on `.terraform.lock.hcl`. `BENCH_TF_ROOT` fixes both: discovery and isolation. This was the last `Path(__file__)`-relative repo-root assumption in the package (judge skills already use `importlib.resources`).

The env read is lazy (per call, not at import), following the `TF_DATA_DIR` precedent in the same module.

## Why not ship tf/ as package data instead?

Complementary, not competing: package data would only fix the *bundled* stacks (`prebuilt/kind` default), while this override is what lets downstream repos point at **their own** stack trees with isolation intact. The bundled default remains unavailable under pip either way until a package-data follow-up; happy to discuss that separately.

## Safety guard

Pointing `BENCH_TF_ROOT` at an ancestor of the run scratch dir would make the per-run whole-tree copy recurse into its own output until the OS path-length limit (reproduced empirically: ~260-deep nesting before `File name too long`). `_isolated_work_dir` now detects that case, refuses the copy with a warning, and degrades to the shared stack dir — provisioning never fails on it.

## Tests

New cases in `tests/unit/deployers/test_deployers_tofu.py`: override resolution, override + `TF_DATA_DIR` isolation (disjoint roots), missing-stack error naming the checked root and the override, blank-override fallback, and the ancestor-guard refusal. A new `tests/unit/deployers/conftest.py` autouse fixture drops ambient `BENCH_TF_ROOT` so a developer's shell cannot skew the suite. The `_TF_ROOT` constant is kept as the default, so existing imports in both deployer test files are untouched.

`uv run pytest tests/unit` (882 passed), `ruff check` / `ruff format --check`, `pre-commit run --all-files`, and `uv lock --check` all clean. Verified from a wheel install in a clean venv: default root points into `site-packages` (the bug), `BENCH_TF_ROOT=/tmp/stacks` resolves a relative stack with no repo checkout.

### DIFF
--- a/devops_bench/deployers/tofu.py
+++ b/devops_bench/deployers/tofu.py
@@ -12,7 +12,13 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-"""OpenTofu-backed deployer driving repo-local ``tf/`` stacks."""
+"""OpenTofu-backed deployer driving ``tf/`` stacks.
+
+Relative stack names resolve under the stack root: ``$BENCH_TF_ROOT`` when set,
+else the checkout's ``<repo_root>/tf``. The override is what lets a
+pip-installed devops-bench (no repo checkout, ``tf/`` is not packaged) drive
+stacks maintained in a downstream repository.
+"""
 
 from __future__ import annotations
 
@@ -23,7 +29,7 @@ import shutil
 from pathlib import Path
 from typing import TYPE_CHECKING, Any
 
-from devops_bench.core import ClusterInfo, ConfigError, get_logger
+from devops_bench.core import ClusterInfo, ConfigError, get_env, get_logger
 from devops_bench.core.subprocess import run
 from devops_bench.deployers.base import Deployer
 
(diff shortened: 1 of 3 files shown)


janetkuo (approved):
/lgtm

PR #54, opened 2026-07-30
feat(verification): add the external_http_probe verifier

geojaz:
This adds an `external_http_probe` verifier that answers a question none of the existing checks can: is the workload actually reachable from outside the cluster?

It discovers the target's LoadBalancer address from the Service or Ingress status via kubectl, then probes it directly from the runner with stdlib urllib: no proxy, no redirect following, and no TLS identity verification, since the probe typically hits a bare LB IP before DNS or certificates exist. It polls through the verifier base's poll helper so a still-provisioning LoadBalancer converges instead of failing, and every I/O call is explicitly bounded: the kubectl discovery call gets `probe_timeout` plus a small fixed overhead constant (a deliberate estimate, called out for review), and the HTTP request gets `probe_timeout`.

Limitations are documented on the class: first match wins on multi-ingress status, no Gateway API discovery yet, and TLS gives reachability only since identity is not verified.

22 unit tests with all I/O mocked. No new dependencies.

This also unblocks the `serving-http` objective in #47, which is commented out until this verifier exists.

### DIFF
--- a/devops_bench/verification/verifiers/external_http_probe.py
+++ b/devops_bench/verification/verifiers/external_http_probe.py
@@ -1,0 +1,244 @@
+# Copyright 2026 The Kubernetes Authors.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""Verifier that GETs a service's external address from the verifier host (off-cluster)."""
+
+from __future__ import annotations
+
+import re
+import ssl
+import urllib.error
+import urllib.request
+from typing import Any, Literal
+
+from pydantic import model_validator
+
+from devops_bench.core import SubprocessError, get_logger
+from devops_bench.k8s import get_resource
+from devops_bench.verification.base import (
+    VERIFIERS,
+    BaseVerifier,
+    VerificationResult,
+    VerificationStatus,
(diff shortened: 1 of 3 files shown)


geojaz:
Flagging something a reviewer might reasonably ask about: this verifier is not pinned to the run's cluster the way the verifiers on main are.

That pinning comes from #79, which adds a `context` field to `BaseVerifier`. The commit applying it to `external_http_probe` is written and tested, but it needs #54, #55 and #79 present at the same time, and no single branch has that combination yet. Splitting it so that just the `external_http_probe` half landed here was tried and fails, because this branch has no `context` field to read.

So it is parked until those merge, then comes in as a small follow-up. Not an oversight, just ordering.

devops_bench/verification/verifiers/external_http_probe.py
```diff
@@ -0,0 +1,244 @@
...
+    kind: str = "service"
+    name: str | None = None
+    selector: str | None = None
+    namespace: str | None = None
+    scheme: Literal["http", "https"] = "http"
+    port: int | None = None
+    path: str = "/"
+    expect_status: int = 200
+    expect_body_matches: str | None = None
+    probe_timeout: int = 10
```
itssimrank, line 159: nit: This should be consistently float type everywhere.

devops_bench/verification/verifiers/external_http_probe.py
```diff
@@ -0,0 +1,244 @@
...
+            return "error", f"unexpected error listing {self.kind}: {exc}", None
+        try:
+            if not objects:
+                return "fail", f"no {self.kind} matched", None
+            target = next(((o, a) for o in objects if (a := self._address(o))), None)
+            if target is None:
+                return "fail", f"no external address on {self.kind} yet", None
+            obj, address = target
+            port = self._resolve_port(obj)
+            url = f"{self.scheme}://{_fmt_host(address)}:{port}{self.path}"
```
itssimrank, line 233: Can we add a ipaddress.ip_address() guard before the url gets constructed?

itssimrank (approved):

PR #59, opened 2026-07-31
feat(verification): implement hold mode for verification entries

geojaz:
This implements `mode: hold`, the third timing mode the entry schema has parsed but rejected since the deterministic verification work landed in #47. A hold entry verifies that a state STAYS true: the check subtree is sampled across a window (default 30s, 5s interval, both overridable per entry) and every sample must pass. Where converge gives a condition time to become true, hold denies it time to stop being true, which is what stability objectives actually mean: a service that stays reachable through a disruption, a restart count that stays flat instead of merely being low at one lucky poll.

Semantics guarantees, all pinned by tests:
- A window that cannot be fully observed never passes vacuously: if the remaining entry budget cannot cover the window, the entry reports `error` before sampling, and a deadline arriving mid-window reports `error` naming the partial observation. The mid-window branch is unreachable through the current `run_entry` path (the upfront guard subsumes it); it stays as a guarded invariant so future callers or refactors cannot silently reintroduce truncation-as-pass.
- An observed failing sample fails the hold immediately, carrying that sample as the child result.
- Evaluation errors do not end sampling but taint the window: no fail plus any errored sample reports `error`, never `pass`, since the condition may have held but was not observed at every sample.
- Samples dispatch through the runner's single-shot path, so compound subtrees compose and leaf I/O stays bounded exactly as assert entries already are.

Hold is explicit-only: objectives still default to converge and safeguards to assert, and declaring the hold window or interval fields on a non-hold entry is a validation error rather than a silent no-op.

11 new unit tests, no real sleeping (scripted clocks). Unblocks the service-uptime and crash-loop stability tasks queued for porting, which declare 60s and 90s holds.

## Summary by CodeRabbit

* **New Features**
  * Added explicit “hold” verification mode for conditions that must remain true throughout a configured time window.
  * Added configurable hold duration and polling interval, with sensible defaults when omitted.
  * Supports repeated checks of compound verification conditions.

* **Bug Fixes**
  * Hold checks now fail immediately when a condition becomes false.
  * Incomplete, errored, or budget-exhausted checks are reported clearly instead of passing.
  * Hold mode no longer activates implicitly based on verification role.

### DIFF
--- a/devops_bench/verification/runner.py
+++ b/devops_bench/verification/runner.py
@@ -78,6 +78,15 @@ _SINGLE_SHOT_WAIT_CEILING_SEC = 120.0
 # path and the single-shot wait-ceiling path in :meth:`VerifierAgent._run_parallel`.
 _PARALLEL_INCOMPLETE_REASON = "evaluation did not complete before the deadline"
 
+# Hold-mode defaults used when an entry sets mode: hold but does not override
+# the window / interval. Explicit-only: resolved_mode never derives "hold"
+# from a role, so these apply only to an entry that opts in by name. These
+# join the timing vocabulary (_MIN_LEAF_BUDGET_SECONDS,
+# _SINGLE_SHOT_WAIT_CEILING_SEC) a future shared timing module will
+# consolidate.
+_DEFAULT_HOLD_WINDOW_SEC = 30.0
+_DEFAULT_HOLD_INTERVAL_SEC = 5.0
+
 
 def _node_name(node: Any) -> str | None:
     """Echo the optional ``name`` label from a spec node, if any."""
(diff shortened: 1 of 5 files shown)


geojaz:
Closing this in favor of #84, which now implements `mode: hold` for both roles.

These two PRs were independent implementations of the same feature. Both branched from the same commit and both removed the same "mode 'hold' is not yet supported" guard, so they were never going to land together. That is my fault for having them open at once.

Working through the difference turned out to be worth it. What this PR did (sample after the agent's turn, in a bounded window) and what #84 did (sample live on a background thread during the turn) are not two implementations of one feature. They are two different features:

- An **objective** starts false and must become true, then stay true. It can only be soaked after the turn ends. That is this PR's model.
- A **safeguard** is an invariant that must never break, including in the middle of the turn where a post-run check cannot see it. That is #84's model.

Both are now in #84, dispatched on `role`. This PR's post-run loop survives there as `run_hold_window`, and its sampling budget behavior carried over with it. Nothing from here is being dropped.

PR #60, opened 2026-07-31
fix(tasks): mint the GitOps repo under an owned scratch root

geojaz:
Follow-up to the safety concern raised on #47 (https://github.com/kubernetes-sigs/devops-bench/pull/47#discussion_r3686958545): setup.sh ran `rm -rf` on an env-derived, tilde-expanded, user-overridable path, so a misconfigured `repo_path` such as `~` would have deleted the operator's home directory.

Rather than guarding the delete with validation, this makes the disaster class unexpressible: mint, do not guard. The GitOps repo now always lives under an owned scratch root (`/tmp/devops-bench` by default on the script side, pinned rather than TMPDIR-derived because the path appears in agent-facing prompt text and must be deterministic; overridable only as a root via `DEVOPS_BENCH_SCRATCH_ROOT`). The old free-form `repo_path` variable is replaced by a validated name-only `repo_name` knob, so callers can influence the leaf name but can never supply a path. The reseed goes through a `safe_remove` helper that asserts the minting contract as a backstop: target strictly under the validated root, `*.git` basename, never the root, `$HOME`, `/`, or a symlink. The root itself is validated too, since an unvalidated root would make every downstream containment check vacuous.

The same pattern ships as a paved path for harness code: `devops_bench.core.scratch` provides `scratch_root` / `mint_dir` / `remove_minted` (containment-asserting), adopted by the agent workspace lifecycle in the harness. AGENTS.md gains a short destructive-operations section and the reviewer configuration now flags destructive operations on derived, unminted paths, so future tasks get steered onto the pattern at authoring and review time.

### DIFF
--- a/.coderabbit.yaml
+++ b/.coderabbit.yaml
@@ -17,7 +17,11 @@ reviews:
   path_instructions:
     - path: "devops_bench/**/*.py"
       instructions: |
-        Review against Python 3.12+ idioms. All Python code must include type hints. Ensure docstrings are provided for public classes and functions. Enforce `uv` for dependency/environment management and `ruff` for linting and formatting. Ensure Apache 2.0 license headers are present on all new source files.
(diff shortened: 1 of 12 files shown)


geojaz:
Heads up: #72 (agent sandbox) has a staged follow-up commit that mounts scratch_root() into the sandbox container read-write at the same path on both sides, and it is waiting on this PR to land first. The home-directory check and the depth guard here are load-bearing for that mount, not just for the GitOps repo minting.

PR #61, opened 2026-07-31
Share the verification budget across converging entries

jessie1111101:
`_run_verification` consumed `VERIFICATION_TOTAL_BUDGET_SEC` first come first served: each
converging entry took `min(timeout_sec, remaining)`. On a task whose objectives fail, the
early entries each poll to the full per-entry cap, and every later entry is recorded
`verification total budget exhausted before evaluation` without ever running.

With the shipped values this is deterministic rather than flaky. 600s total and a 120s
per-entry cap evaluate five entries and abandon everything after them.

## Why it matters beyond lost coverage

An abandoned entry counts toward neither the numerator nor the denominator in `rollup`, so
the run still emits a correctness score. It is just computed from whichever entries
happened to run first, and it looks like a normal result.

Found this on a live run of `deploy-hello-app`, which declares 13 objectives:

```
status counts: {'fail': 4, 'error': 9, 'pass': 2}
  8 of the 9 errors: "verification total budget exhausted before evaluation"

VerificationCorrectness  0.0     <- computed from 4 of 13 objectives
VerificationCoverage     0.4
OutcomeScore             0.0
```

`VerificationCoverage` correctly flagged that only 40% of checks were evaluated, but the
correctness score alongside it carries no such warning.

## The change

The converging entries now share what remains of the total, each getting
`min(timeout_sec, remaining / converging_entries_left)`. The share is recomputed per entry
from the live remaining time, so an entry that finishes early hands its unused budget back
to the ones after it rather than forfeiting it. Assert entries stay out of the divisor,
since they evaluate once and consume no budget.

The total cap is unchanged, so the worst case wall clock for the pass does not grow. For
the 13-entry task above, every entry now gets a real budget (46s for the first, rising as
earlier entries return) instead of five entries taking everything.

Raising the cap was the other option, but it does not scale: guaranteeing 13 entries at
120s each needs 26 minutes per task, and the next task with 20 objectives breaks again. The
comment above the constant already explains why the cap exists.

## Tests

Two regression tests. The first drives 10 converging entries against a 100s total and
asserts none is abandoned and the first does not take the full per-entry cap; I confirmed
it fails against the old behavior. The second asserts an early finisher's unused share is
redistributed rather than lost.

/kind bug

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved verification scheduling so checks share available time fairly.
  * Redistributed unused time from early-finishing checks to later checks.
  * Preserved converging-check time when independent assertions run slowly.
  * Applied minimum verification time and clearer error reporting when convergence is insufficiently observed.

* **Tests**
  * Added coverage for time allocation, redistribution, assertion timing protection, and failure classification.

### DIFF
--- a/devops_bench/evalharness/default.py
+++ b/devops_bench/evalharness/default.py
@@ -448,28 +448,41 @@ class DefaultEvalHarness(Harness):
         evaluator.
 
         Two budgets apply. ``timeout_sec`` is the per-entry cap for a single
-        converging entry's checks. :data:`VERIFICATION_TOTAL_BUDGET_SEC` is
-        the wall-clock cap for this whole pass across every entry; without it
-        a task with many failing converge objectives burns entries x
-        ``timeout_sec`` (12 entries x 120s is 22+ minutes). A single monotonic
-        deadline is computed from the total budget once at the top, and the
-        converging entries **share** what remains of it: each gets
-        ``min(timeout_sec, remaining / converging_entries_left)``. Sharing is
-        what keeps the total cap from being consumed first-come-first-served,
+        converging entry's checks. :data:`VERIFICATION_TOTAL_BUDGET_SEC`
+        bounds the converging entries as a group; without it a task with many
+        failing converge objectives burns entries x ``timeout_sec`` (12
+        entries x 120s is 22+ minutes). It is not a cap on the pass as a
+        whole: each assert entry runs outside the budget and pushes the
+        deadline out by its own duration, so the worst case is the total
+        budget plus the sum of those durations, and an assert leaf floors its
+        own kubectl call rather than inheriting a zero budget (see
+        :func:`~devops_bench.verification.base.single_call_timeout`).
+
+        A single monotonic deadline is computed from the total budget once at
+        the top, and the converging entries **share** what remains of it: each
+        gets ``min(timeout_sec, remaining / converging_entries_left)``. Sharing
+        is what keeps the total cap from being consumed first-come-first-served,
         where a handful of early entries polling to their own cap leave the
-        rest unevaluated and silently drop out of the score's denominator.
-        The share is recomputed from the live remaining time, so an entry that
-        finishes early hands its unused budget back to the ones after it, and
-        the deadline is pushed out by however long each assert entry took so a
-        slow single evaluation does not shorten the converging entries after
-        it. Assert entries
-        ignore the total budget and always run: they are single evaluations,
-        and a safeguard that goes unchecked defeats the point of having it.
-        A converging entry with less than :data:`MIN_LEAF_BUDGET_SECONDS`
-        remaining is recorded here as budget-exhausted rather than handed to
-        ``run_entry``: the runner's own leaf guard uses that same threshold
-        to short-circuit an under-budget leaf as a definite "deadline
-        exhausted" outcome, and this entry was never observed either way.
+        rest unevaluated and silently drop out of the score's denominator. The
+        share is recomputed from the live remaining time, so an entry that
+        finishes early hands its unused budget back to the ones after it.
+        Assert entries ignore the total budget and always run: they are single
+        evaluations, and a safeguard that goes unchecked defeats the point of
+        having it.
+
+        A converging entry whose share came in under ``timeout_sec`` and did
+        not converge is recorded ``"error"``, not ``"fail"``. Such an entry
+        fails only by reaching a deadline, and a truncated deadline is one this
+        pass imposed rather than one the task agreed to: the condition was not
+        observed false, it was not observed. Scoring it as a miss would trade
+        the old symptom (correctness computed from a fraction of the
+        objectives, with coverage visibly low) for a subtler one: full-looking
+        coverage over failures the harness never saw. A converging entry with
+        less than :data:`MIN_LEAF_BUDGET_SECONDS` remaining is likewise
+        recorded as budget-exhausted rather than handed to ``run_entry``: the
+        runner's own leaf guard uses that same threshold to short-circuit an
+        under-budget leaf as a definite "deadline exhausted" outcome, and this
+        entry was never observed either way.
 
         Args:
             entries: The task's parsed verification entries.
@@ -483,14 +496,22 @@ class DefaultEvalHarness(Harness):
         agent = VerifierAgent()
         report: list[dict[str, Any]] = []
         total_deadline = time.monotonic() + VERIFICATION_TOTAL_BUDGET_SEC
-        # Converging entries share what is left of the total budget evenly, so an
+        # How many converging entries remain from each position onward, so an
         # early entry that polls to its own cap cannot starve the ones after it.
-        # Only converging entries count toward the divisor: assert entries
-        # evaluate once and consume no budget, so including them would shrink
-        # everyone's share for nothing.
-        converging_left = sum(1 for e in entries if e.resolved_mode != "assert")
+        # Counted per position rather than decremented as the loop goes: a
+        # running counter goes stale the moment an entry is skipped before
+        # reaching the decrement, and obliges every mode added later to maintain
+        # it. Assert entries are excluded because they consume no budget, so
+        # counting them would shrink everyone's share for nothing.
+        converging_left: list[int] = []
+        still_to_come = 0
+        for entry in reversed(entries):
+            if entry.resolved_mode != "assert":
+                still_to_come += 1
+            converging_left.append(still_to_come)
+        converging_left.reverse()
 
-        for entry in entries:
+        for index, entry in enumerate(entries):
             remaining = total_deadline - time.monotonic()
             if entry.resolved_mode != "assert" and remaining < MIN_LEAF_BUDGET_SECONDS:
                 # Never evaluated, not a condition observed false.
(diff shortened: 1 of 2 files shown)


geojaz:
Heads up on a cross-PR interaction rather than a problem with this change on its own.

#84 adds a `hold` mode for verification entries, and it takes an early `continue` in the same `_run_verification()` loop this PR touches, before reaching the point where the converging counter gets decremented.

The divisor here counts every entry whose resolved mode is not `assert`:

```python
converging_left = sum(1 for e in entries if e.resolved_mode != "assert")
```

That predates hold mode, so hold entries get counted into the initial total but never decrement it. The effect is that each genuinely converging entry receives a smaller budget than it should, in proportion to how many hold entries the task declares. There is no error and nothing in the output says so. It just looks like a check that gave up early and reported a real failure, when what actually happened was a shrunken budget.

Whichever of these two lands second will need to exclude "hold" from that divisor. Flagging it now so it does not get found at merge time. Happy to handle it on my side if this one goes in first.

jessie1111101:
Confirmed against your diff. Both hold branches `continue` before `converging_left` is decremented, and my divisor is `!= "assert"`, so hold entries are counted in and never come out. Every converging entry gets too small a share and the tail is worst hit, with the last one getting `remaining / (H + 1)` instead of all of it, so budget is both misallocated and left unspent.

Taking your offer: if this lands first, please exclude hold from the divisor on your side.

One suggestion. The divisor being a predicate over modes means every new mode has to remember to opt out, and nothing fails loudly when it does not. Partitioning entries before the loop into early handled versus budget sharing would make it correct by construction. Your two hold branches are already the natural split.

tests/unit/evalharness/test_verification_wiring.py
```diff
@@ -339,3 +339,67 @@ def test_resolve_spec_placeholders_recurses_through_nested_entries() -> None:
...
+
+    with patch(
+        "devops_bench.evalharness.default.VerifierAgent.run_entry", side_effect=fake_run_entry
+    ):
+        _harness()._run_verification(entries, timeout_sec=120)
+
+    # Each returns instantly, so the later shares should not shrink below the
+    # first one's — the unspent remainder is redistributed, not lost.
+    assert len(seen) == 4
+    assert seen[-1] >= seen[0] - 1.0
```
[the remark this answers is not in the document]
jessie1111101: Already addressed before this comment landed: the assertion is `assert seen[-1] > seen[0]`, which a static per entry share would not satisfy.

devops_bench/evalharness/default.py
```diff
@@ -495,8 +510,17 @@ def _run_verification(
...
+            else:
+                # Recomputed from the live remaining time, so an entry that
+                # finishes early hands its unused share back to the rest.
+                entry_budget = max(remaining / converging_left, MIN_LEAF_BUDGET_SECONDS)
+                converging_left -= 1
+
+            started = time.monotonic()
             try:
-                result = agent.run_entry(entry, timeout_sec=min(timeout_sec, remaining))
+                result = agent.run_entry(entry, timeout_sec=min(timeout_sec, entry_budget))
```
isadominguez314: I ran into this problem with my testing. With some checks, the per check budget can be too low (in deploy-hello-app it is ~46s per check). In some cases, this isn't enough time to finish the task and something that gets interrupted due to not enough time is incorrectly marked as failed. More details:

The status="fail" default in _failed (runner.py:87-94) is documented as "a deadline that expires without the condition converging is a definite fail." That was written when every converging entry received the full per-entry cap, waiting the whole agreed 120s and observing nothing is fairly called a failure. Sharing the budget invalidates the premise: for the 13-objective task in the description the first entry gets 46s, and a rollout or a cold image pull can plausibly need more than that. It's now recorded as a definite failure and counted in the objective denominator (rollup.py:110).

The part that worries me most is that the signal which used to catch this is gone. Under the old behavior an under-served entry became error, dropped out of both numerator and denominator, and VerificationCoverage fell to 0.4 — the run announced that it was only 40% measured. Now the same entry becomes fail, drags VerificationCorrectness down, and coverage reads 1.0. The description rightly criticizes the old path for producing "a plausible-looking correctness value"; I think this trades that for a differently-wrong value that is harder to notice.

Would it work to record a truncated timeout as error rather than fail? The call site already knows whether it truncated — entry_budget < timeout_sec — and error means precisely "could not be evaluated, not observed false" (base.py:96-100). That keeps everything this PR wins: every entry is reached, and a spec whose entries mostly pass fast now scores in full. It just stops the harness from asserting a failure it did not actually observe, and keeps coverage honest about runs that were genuinely under-budgeted.

The trade is that a truly-failing truncated entry also becomes error instead of a scored miss. That seems like the right direction to err for a benchmark. Either way it's worth being deliberate, because assert report[0]["status"] == "fail" at test_verification_wiring.py:379 pins the current behaviour for an entry that received 10s of a 120s cap.

devops_bench/evalharness/default.py
```diff
@@ -495,8 +510,17 @@ def _run_verification(
                 )
                 continue
 
+            if entry.resolved_mode == "assert":
+                entry_budget = remaining
+            else:
+                # Recomputed from the live remaining time, so an entry that
+                # finishes early hands its unused share back to the rest.
+                entry_budget = max(remaining / converging_left, MIN_LEAF_BUDGET_SECONDS)
+                converging_left -= 1
```
isadominguez314: The decrement lives in the else branch, but the budget-exhausted path at line 511 continues before reaching it, so a skipped entry stays in the divisor permanently.

On its own that's harmless — once the budget is gone everything after is skipped too. It becomes reachable because of the new finally at 538: an assert entry pushes total_deadline forward, so remaining can climb back above MIN_LEAF_BUDGET_SECONDS and a later converging entry runs against a stale divisor.

entries: [converge A, converge B, assert S, converge C]   # divisor starts at 3
A runs, nearly exhausts the budget           -> converging_left 2
B: remaining < 1.0s, skipped, no decrement   -> converging_left 2   (stale)
S: assert, takes 5s                          -> total_deadline += 5
C: remaining ~5s, divisor 2                  -> 2.5s instead of the full 5s
C is silently under-served by half, which feeds straight into the truncation problem above.

Your partition-before-the-loop suggestion would fix this. A smaller alternative that gets the same correctness-by-construction property is a static per-index suffix count instead of a running decrement — a position-based count can't go stale, and it removes the "every new mode must remember to decrement" hazard that hold already ran into:

converging_left: list[int] = []
still_to_come = 0
for entry in reversed(entries):
    if entry.resolved_mode != "assert":
        still_to_come += 1
    converging_left.append(still_to_come)
converging_left.reverse()
then remaining / max(converging_left[index], 1) in the loop. Skipped entries simply aren't in any later entry's count, since the count only looks forward from each position.

devops_bench/evalharness/default.py
```diff
@@ -452,8 +452,17 @@ def _run_verification(
         the wall-clock cap for this whole pass across every entry; without it
```
isadominguez314: This sentence isn't always true anymore. The `finally` at line 538 pushes the deadline out by each assertion's duration so the actual bound is `VERIFICATION_TOTAL_BUDGET_SEC` plus the sum of all assert entry durations. Each assert leaf floors its kubectl call at 30s (base.py:66-72), so a spec with several assert safeguards against a slow or wedged API server can add minutes to the worst case.

devops_bench/evalharness/default.py
```diff
@@ -495,8 +510,17 @@ def _run_verification(
                 )
                 continue
 
+            if entry.resolved_mode == "assert":
+                entry_budget = remaining
```
isadominguez314: I think this is dead. `run_entry` sets `deadline = time.monotonic() + (0.0 if single_shot else timeout_sec) so timeout_sec is discarded entirely for assert entries. Harmless but reads as though assert gets a budget which is confusing. You could drop the branch and pass the converging budget unconditionally, or add a note that the value is ignored downstream.

devops_bench/evalharness/default.py
```diff
@@ -495,8 +510,17 @@ def _run_verification(
...
+            else:
+                # Recomputed from the live remaining time, so an entry that
+                # finishes early hands its unused share back to the rest.
+                entry_budget = max(remaining / converging_left, MIN_LEAF_BUDGET_SECONDS)
```
isadominguez314: This floor isn't protecting against anything. The guard at 507 already guarantees remaining >= MIN_LEAF_BUDGET_SECONDS before we get here.

As written it lets an entry take more than its share when remaining / converging_left drops below 1.0s, which re-creates tail starvation in miniature: the early entries each get their 1.0s and the last ones fall off into the budget-exhausted path. At the shipped constants you'd need on the order of 600 entries for this to bite, so it's theoretical — but the neighbouring lines all carry a rationale and this one doesn't, so a future reader can't tell whether the overspend is intentional.

tests/unit/evalharness/test_verification_wiring.py
```diff
@@ -339,3 +339,115 @@ def test_resolve_spec_placeholders_recurses_through_nested_entries() -> None:
...
+    monkeypatch.setattr("devops_bench.evalharness.default.VERIFICATION_TOTAL_BUDGET_SEC", 10.0)
+
+    slow_assert_sec = 1.0
+    seen: list[tuple[str, float]] = []
+
+    def fake_run_entry(entry: object, timeout_sec: float = 120) -> VerificationResult:
+        name = entry.name  # type: ignore[attr-defined]
+        seen.append((name, timeout_sec))
+        if name == "safeguard":
+            time.sleep(slow_assert_sec)
```
isadominguez314: nit: This puts a real one-second sleep in the unit suite. Since the production code reads the clock via `time.monotonic()` in `_run_verification`, injecting a fake clock (as `poll_until` already allows for) would make the test both faster and deterministic — the current 0.1s and 0.5s tolerances in the assertions below are absorbing real scheduler jitter.

isadominguez314 (changes requested):

devops_bench/evalharness/default.py
jessie1111101: Agreed, done. A converging entry that gets less than the full `timeout_sec` and comes back `fail` is now recorded `error`.

I tried first to reclassify only entries that actually ran out of time, but `_failed` sets `elapsed_time=0.0` (runner.py:99), so there is nothing to tell "ran out" from "answered false fast." Your rule works because the call site knows `granted < timeout_sec` before it calls, and it only applies to converging entries.

The reason says so: `not observed: given 10.0s of the 120s converge budget; <original reason>`. The test at :379 now pins `error` instead of `fail`.

devops_bench/evalharness/default.py
jessie1111101: Taken as written, suffix counts and all.

One note: I could not get it to actually bite. The `finally` adds back exactly the assert's own duration, and the clock advanced by that same amount while it ran, so `remaining` is restored, never raised. It only ever decreases, so once one entry is skipped every later one is too and C never runs.

Still worth fixing. The counter breaks the moment a future `continue` lands above the decrement, and it makes every new mode responsible for maintaining it. Your version cannot go stale. I kept the `max(..., 1)`: when the tail is all assert entries the count is 0, so it is load bearing.

devops_bench/evalharness/default.py
jessie1111101: Fixed. It now says the budget bounds the converging entries as a group, not the pass, and that the worst case is the budget plus the sum of the assert durations. Added your 30s floor point with a pointer to `single_call_timeout`.

devops_bench/evalharness/default.py
jessie1111101: Right, dropped. `run_entry` computes `deadline = time.monotonic() + (0.0 if single_shot else timeout_sec)`, so an assert entry never sees the value. The budget is passed unconditionally now, with a one line note that it is ignored downstream. It also removes a branch the truncation check above would have had to straddle.

devops_bench/evalharness/default.py
jessie1111101: Correct, it is not a guard. I kept it and wrote down why: a leaf handed a fraction of a second returns nothing useful, so I would rather it take its full second and let the tail fall into the budget exhausted path, where it is now recorded as not observed. The comment names the cost and the point where it starts to matter, around `VERIFICATION_TOTAL_BUDGET_SEC / MIN_LEAF_BUDGET_SECONDS` entries.

tests/unit/evalharness/test_verification_wiring.py
jessie1111101: Done. Added a `_FakeClock` the test advances by hand, patched over the module's `time`. The sleep is gone and so are the tolerances: the three budget tests now assert exact shares (`100.0 / 10`, `[40/4, 40/3, 40/2, 40/1]`, `10.0`).

I only touched the sleep this PR added plus the two other new tests. The three older ones, now at lines 177, 213 and 240, predate this change so I left them.

tests/unit/evalharness/test_verification_wiring.py
```diff
@@ -339,3 +357,124 @@ def test_resolve_spec_placeholders_recurses_through_nested_entries() -> None:
...
+        r["reason"] == "verification total budget exhausted before evaluation" for r in report
+    )
+    # The first entry got exactly its fair share (100/10), not the full 120s cap.
+    assert seen[0] == 100.0 / 10
+    # ...and because that share is a fraction of the 120s the task agreed to, a
+    # non-convergence is not an observation that the condition is false. Scoring
+    # it "fail" would report full coverage over a failure never actually seen.
+    assert report[0]["status"] == "error"
+    assert report[0]["success"] is False
+    assert "not observed" in report[0]["reason"]
```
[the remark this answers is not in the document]
jessie1111101: The intent is right but the literal patch fails: `_FakeClock` only moves on `advance()`, and the share is `remaining / converging_left[index]`, so a frozen clock gives `100/10, 100/9, 100/8` and so on. Result is `At index 1 diff: 11.11111111111111 != 10.0`. The neighbouring test bakes that growth in as `[40/4, 40/3, 40/2, 40/1]`.

Took the stronger version instead: the fake now advances the clock by the timeout it was granted, modeling the case the regression is about, an entry that never converges and polls out its share. With the budget actually depleting, all ten shares are `100/10` and sum to exactly 100s, so the assertion is both true and meaningful. Also added `clock.now == 100.0` and generalised the three status assertions to all ten entries. 20 passed.

PR #64, opened 2026-07-31
Add the prebuilt/minimum TF stack

jessie1111101:
`tasks/gcp/deploy-hello-app/task.yaml` declares `stack: "prebuilt/minimum"`, but `tf/prebuilt/minimum/` is not in this repo — `git log upstream/main -- tf/prebuilt/minimum` is empty, so it never landed.

The effect is that the only GCP task here cannot run at all. `TFDeployer.__init__` raises `ConfigError: TF stack not found under <tf_root>: prebuilt/minimum` before any cluster is created.

In `migrated.bara.sky` the task and the stack are a single flip-group:

```
# "tasks/gcp/deploy-hello-app/**",  # flip-group: minimum
# "tf/prebuilt/minimum/**",  # flip-group: minimum
```

The task came over with #47 and the stack did not follow.

## The port

Unmodified apart from the Apache headers. The stack consumes `tf/modules/cluster` via `infra_provider` / `cluster_name` / `location` / `node_count` / `machine_type` / `project_id` / `kubeconfig_path`, and this repo's module already exposes exactly that variable set (identical to gke-labs', including the provider-neutral abstraction), so nothing needed adapting.

It also carries a destroy-time `null_resource` that sweeps the `hello-app-<cluster>` Artifact Registry repo, which the agent creates project-globally and outside this stack's state — without it, teardown leaks a repo per run.

## Verification

- `TFDeployer(tf_dir="prebuilt/minimum")` now resolves instead of raising.
- `tofu fmt -check` clean; `hack/boilerplate.py --dry-run` clean.
- Not applied against a live project from this checkout. This exact stack has been running the task on our bastion, which is how the gap surfaced: it had been hand-placed there, masking the fact that it was never migrated.

## Interaction with #63

#63 adds `docs/appendix/known_issues.md` with a router row describing this break, and drops the Artifact Registry hacks-table row on the grounds that the stack was absent. Once this merges, that pairing should flip — the router row goes away and the AR-sweep row comes back. I will send that follow-up rather than fold it in here, since the two PRs touch different areas and either could land first.

### DIFF
--- a/tf/prebuilt/minimum/main.tf
+++ b/tf/prebuilt/minimum/main.tf
@@ -1,0 +1,93 @@
+# Copyright 2026 The Kubernetes Authors.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+terraform {
+  required_version = ">= 1.5.0"
+  required_providers {
+    google = {
+      source  = "hashicorp/google"
+      version = ">= 5.0.0"
+    }
+    null = {
+      source  = "hashicorp/null"
+      version = ">= 3.0.0"
+    }
+    kind = {
+      source  = "tehcyx/kind"
+      version = ">= 0.5.0"
+    }
+  }
+}
+
+provider "google" {
+  project = var.project_id != "" ? var.project_id : null
+  region  = var.location != "" && var.location != "local" ? var.location : null
+}
+
+# Sweep the Artifact Registry repo the deploy-hello-app agent creates
+# (hello-app-<cluster>, project-global and NOT in this stack's tofu state) so it
+# doesn't leak across runs — the cluster teardown alone never removes it. No-op
+# for the other tasks on this stack (the repo won't exist).
+#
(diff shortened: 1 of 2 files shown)


geojaz (approved):
/lgtm

I don't have permissions for real, but hopefully providing a useful +1 signal for approvers.

jessie1111101:
Heads up: this is necessary but not sufficient. With the stack restored, `deploy-hello-app` fails one step later with `stack 'prebuilt/minimum' requires an explicit provider` — the task also lost its `provider: "gcp"` line in migration. #70 fixes that. Both are needed before the task can run.

Confirmed by running it end to end: with both applied it provisions GKE and reaches the agent.

geojaz:
> Heads up: this is necessary but not sufficient. With the stack restored, `deploy-hello-app` fails one step later with `stack 'prebuilt/minimum' requires an explicit provider` — the task also lost its `provider: "gcp"` line in migration. #70 fixes that. Both are needed before the task can run.
> 
> Confirmed by running it end to end: with both applied it provisions GKE and reaches the agent.

+1- let's pls get this merged; I don't have permissions to actually lgtm 😇 but @jessie1111101 is correct- and I think she's fixing a bug I introduced.

janetkuo (approved):
/lgtm

tf/prebuilt/minimum/main.tf
```diff
@@ -0,0 +1,93 @@
...
+resource "null_resource" "ar_cleanup" {
+  triggers = {
+    project = var.project_id
+    repo    = "hello-app-${var.cluster_name}"
+  }
+
+  provisioner "local-exec" {
+    when    = destroy
+    command = <<-EOT
+      gcloud artifacts repositories list --project='${self.triggers.project}' \
```
janetkuo, line 58: While tf/ is provider-scoped, this calls gcloud and creates a minor leak when running the stack locally on KinD. Please add a guard.

PR #69, opened 2026-07-31
Fix the opa-remediation PolicyReport readiness gate

jessie1111101:
`opa-remediation` fails end to end on kind, every time, before the agent starts:

```
ERROR: PolicyReports never showed failing results for all seeded workloads after 180s
Error: local-exec provisioner error
```

> [!NOTE]
> This PR originally proposed only the timeout and diagnostics changes I suggested in [#47 (comment)](https://github.com/kubernetes-sigs/devops-bench/pull/47#discussion_r3692726781), on the theory that 180s was too thin for a cold cluster. **That theory was wrong**, and I have rewritten this description to match what the fix actually is. Raising the timeout would not have helped: the gate could never pass at any duration.

## Root cause

The gate collected `(policy, resource)` pairs from `results[].resources`. Kyverno 1.12 emits **one PolicyReport per subject** and names that subject in the report's top-level `scope`; `results[]` entries carry **no `resources` field at all**. So the inner loop always iterated an empty sequence, the pair set stayed permanently empty, and the wait always timed out claiming all four workloads were missing — regardless of how long it waited.

Confirmed against a live kind cluster, re-run with `--no-teardown`. The seeded workloads had failing results the whole time:

```
scope:   {'kind': 'Deployment', 'name': 'web', 'namespace': 'team-alpha', ...}
summary: {'error': 0, 'fail': 1, 'pass': 1, 'skip': 0, 'warn': 0}
results[0]: {'policy': 'disallow-privileged-containers', 'result': 'pass', ...}   # no 'resources' key
```

Matching on `scope` satisfies all four required pairs immediately; matching on `results[].resources` yields the empty set.

## What changed

1. **Read the subject from the report `scope`** when `results[].resources` is absent — the actual fix. Only `Deployment`-scoped reports are matched; the ReplicaSet and Pod reports for the same workload carry generated names (`cache-6489fdb6f9`) that never equal the seeded ones. The listed-resources path is kept for report shapes that populate it.
2. **180s → 600s.** Now headroom rather than a fix: with the gate working, reports arrive in seconds.
3. **The failure names the missing pairs.** The old message could not distinguish three-of-four from zero-of-four — which is exactly why this was misread as a timeout for as long as it was.
4. **A stalled scan fails early.** If the missing set stops shrinking for 90s, exit with those pair names instead of burning the remaining budget.

Points 3 and 4 are what made the root cause findable, so they stay.

## Verification

Re-ran the task end to end on the bastion against a fresh kind cluster:

```
==> Waiting for Kyverno's background scan to populate PolicyReports...
    PolicyReports populated with violations for all seeded workloads.
```

The embedded check was also exercised directly against the live cluster's `kubectl get policyreport -A -o json` (ready, nothing missing), against empty input (all four reported missing), and against malformed input (exit 2). `bash -n` and `hack/boilerplate.py --dry-run` clean.

### DIFF
--- a/tf/prebuilt/opa-remediation/scripts/setup.sh
+++ b/tf/prebuilt/opa-remediation/scripts/setup.sh
@@ -94,12 +94,16 @@ echo "==> Waiting for Kyverno's background scan to populate PolicyReports..."
 # already compliant. Wait until failing results collectively name the exact
 # four (policy, resource) pairs seeded, not just the four resource names: a
 # resource name flagged by the wrong policy must not read as ready.
-reports_ready=false
-for _ in $(seq 1 36); do
-  if kubectl get policyreport -A -o json 2>/dev/null \
-       | python3 -c '
+# Prints the still-missing "policy/resource" pairs, one per line, and exits
+# non-zero if the cluster could not be queried at all. Empty output means ready.
+_missing_report_pairs() {
+  kubectl get policyreport -A -o json 2>/dev/null \
+    | python3 -c '
 import sys, json
-data = json.load(sys.stdin)
+try:
+    data = json.load(sys.stdin)
+except ValueError:
+    sys.exit(2)
 pairs = {
     (r.get("policy"), res.get("name"))
     for it in data.get("items", [])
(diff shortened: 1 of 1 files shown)


jessie1111101:
Closing this — the fix already landed on main by another route.

`tf/prebuilt/opa-remediation/scripts/setup.sh` on main now carries the readiness gate this PR
was written to add: the 600s ceiling (`MAX_ATTEMPTS=120` at `POLL_INTERVAL=5`), the
`STALL_ATTEMPTS=18` early-exit for a scan that has stopped making progress, and the diagnostic
that names the missing policy/resource pairs. That arrived via #47 and #66 rather than here.

The whole remaining delta between this branch and main is a single line:

```
kubectl -n kyverno wait --for=condition=Available deploy --all --timeout=300s
```

which would *replace* main's retry loop around that wait with a one-shot version — strictly
weaker than what is there now. So there is nothing left to merge, and merging it would be a
small regression.

Nothing is lost by closing: the root cause and its fix are on main, and the reviewer feedback
here was about the superseded version of the change.

PR #72, opened 2026-08-01
feat(agents): optionally run CLI agents in a container with a scoped cluster identity

geojaz:
Agent transcripts showed CLI agents ranging well outside the task: reading files from
the operator's home directory, copying archives out of ~/Downloads, running `rm -rf`,
and in one case running `find ~ | grep -i <task-name>` to hunt for the benchmark's own
fixtures. On a workstation checkout that is the harness author's whole filesystem, and
any answer material sitting on disk is reachable.

This adds an opt-in container sandbox for CLI agents. When `BENCH_AGENT_SANDBOX` is set
to `docker`, `1`, or `true`, the agent argv is wrapped in `docker run` with:

* only a per-run scratch workspace and a generated kubeconfig mounted, so the host
  filesystem and the operator's environment are not visible
* `HOME` pointed at the workspace, so agent config and history stay inside the run
* the container joined to the `kind` network with the kubeconfig server rewritten to the
  control-plane node, so TLS still verifies against the existing cert
* credentials passed explicitly rather than by scraping os.environ for well-known names

`BENCH_AGENT_IMAGE` names the image. `hack/agent-sandbox.Dockerfile` builds a suitable
one (node + kubectl + the gemini CLI). `hack/agent-credential.sh` mints a credential for
the agent: by default an API key restricted to generativelanguage.googleapis.com, or a
short-lived impersonated token with `--vertex-sa`, plus `--revoke` to clean up.

Off by default so it can be A/B'd against current behaviour while tasks are still being
debugged.

Worth being explicit about scope: this is filesystem and environment isolation only. It
does not scope what the agent can do to the cluster, which is a separate concern best
handled with per-task RBAC, and it does not help if answer material is seeded inside the
cluster itself.

### DIFF
--- a/devops_bench/agents/cli/gemini_cli/agent.py
+++ b/devops_bench/agents/cli/gemini_cli/agent.py
@@ -50,7 +50,7 @@ from devops_bench.agents.shared.cli_capabilities import (
 )
 from devops_bench.core import SubprocessError, get_logger
 from devops_bench.core.model_providers import resolve_provider
-from devops_bench.core.subprocess import run
+from devops_bench.core.subprocess import CompletedProcess, run
 
 if TYPE_CHECKING:  # pragma: no cover - typing-only import
     from devops_bench.agents.capabilities import McpBinding
@@ -255,9 +255,7 @@ class GeminiCliAgent(AgentHarness):
             sandboxed = False
             if sandbox.sandbox_enabled():
                 cluster = sandbox.current_cluster_name()
-                kubeconfig = (
-                    sandbox.build_agent_kubeconfig(cluster, workdir) if cluster else None
-                )
+                kubeconfig = sandbox.build_agent_kubeconfig(cluster, workdir) if cluster else None
                 if kubeconfig is None:
                     # Refuse rather than silently running unsandboxed on the host: a
                     # containment control that quietly degrades is worse than none.
@@ -270,6 +268,8 @@ class GeminiCliAgent(AgentHarness):
                 )
                 sandboxed = True
 
+            completed: CompletedProcess | None = None
+            timeout_exc: SubprocessError | None = None
             try:
                 completed = run(
                     run_argv,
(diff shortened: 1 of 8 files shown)


geojaz:
Pushed three follow-ups that belong with this change rather than in a separate PR, since they are all sandbox and agent-turn lifecycle:

- Capture the trajectory incrementally, so a run that times out still keeps the record of what the agent did. Previously a timeout discarded the whole trajectory, which threw away exactly the evidence needed to understand the timeout.
- Treat a failed agent turn as a failed run. A turn that errored was still being counted as a successful run, which quietly inflated results.
- Reap the sandbox container on every exit path. The container was leaking on the error and timeout paths, so a run that failed left a container behind holding its scoped cluster identity.

geojaz:
Noting a related follow-up that is deliberately not in this PR, plus one thing worth knowing about this branch.

A fix that expands `${CLUSTER_NAME}` in verification and chaos spec values depends on the container reaping commit here, and separately on the hold mode work in #84. This branch does not carry hold mode, so the commit does not apply to it cleanly today. It is parked until both this PR and #84 land, then it comes in as a small follow-up.

More generally, this branch has drifted from the line the rest of the verification work is on. It has no hold mode, and its `test_verification_wiring.py` is roughly ninety lines shorter than the version the other PRs are built against. Whichever of this PR and #84 merges second should expect a real reconciliation rather than a clean merge. Calling it out now so it is not a surprise at merge time.

geojaz:
Note for reviewers: there is a staged follow-up commit for this PR that bind-mounts the harness scratch root into the sandbox container, read-write, at the same path inside and outside. That lets a task seed a host-side artifact (for example a bare git repo) that both the sandboxed agent and the host-side verifiers reach at one path. It depends on scratch_root() from #60, so I am holding it until #60 merges and this branch picks that up. Keep it in mind when reviewing the mount set here: the workspace-plus-kubeconfig-only posture in this diff will grow by one deliberate mount, and the owned-root guards in #60 are what make a read-write mount safe.

isadominguez314:
The sandbox stack this explored is now landing for real: #181 (core seam + docker executor + kind network plan), #182 (scoped credentials, provider network plans, pod security), #183 (keyless Vertex via a metadata-server emulator). The decisions proven out here and in the gke-labs testbed — exec-plugin detection, fixture mounts as part of the boundary, the admin-token escape hatch — were lifted into the harness's own seams rather than merged as-is; credit and thanks, @geojaz. I'd suggest closing this one in favor of the stack, but that's your call.

PR #76, opened 2026-08-06
feat(providers): implement vcluster provider and OpenTofu deployer integration

isadominguez314:
### Summary
Implements vCluster Provider and OpenTofu Deployer Integration.

> [!NOTE]
> **Stacked PR**: This PR is stacked on top of PR #66. Please review and merge #66 first.

### Changes
- Adds `VClusterProvider` with inline no-op account credentials, a two-step local allowlist check, and secure kubeconfig writing (`0600` permissions) / PV cleanup.
- Updates `TFDeployer` to cache `ClusterInfo`, pass unwrapped outputs to `ensure_cluster_credentials()`, and call `provider.cleanup()` on destroy.
- Updates base `Provider`, `GcpProvider`, and `KindProvider` signatures for compatibility.
- Registers the provider in `PROVIDERS`.

### Verification
- Comprehensive unit tests added for provider logic and verified locally.
- Verified end-to-end execution of a remote benchmark (`opa-remediation`) using this provider via the CLI harness.

## Summary by CodeRabbit

- **New Features**
  - Added support for deploying virtual Kubernetes clusters.
  - Added configurable host context, kubeconfig path, and service exposure options.
  - Added secure kubeconfig generation and cluster connection details.
  - Added support for NodePort and LoadBalancer connectivity.
- **Bug Fixes**
  - Improved cleanup after failed or incomplete cluster teardown.
  - Added validation for invalid cluster output and connection configurations.
  - Improved setup reliability with bounded retries and clearer failure detection.
- **Tests**
  - Expanded coverage for provisioning, credentials, cleanup, and configuration validation.


tf/prebuilt/vcluster/values.yaml.tftpl
```diff
@@ -0,0 +1,81 @@
...
+      enabled: true
+  fromHost:
+    storageClasses:
+      enabled: true
+    nodes:
+      enabled: true
+      selector:
+        all: true
+
+serviceCIDR: "34.118.224.0/20"
```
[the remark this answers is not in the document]
isadominguez314: Loft Labs' Helm chart version 0.20.0 defines serviceCIDR as a top-level property in values.schema.json (line 3157). Moving it under networking.serviceCIDR is required in newer vCluster CLI configuration schemas, but fails Helm schema validation for the 0.20.0 chart version.

devops_bench/providers/vcluster.py
```diff
@@ -0,0 +1,417 @@
...
+        kubeconfig_val = outputs.get("kubeconfig") if outputs else None
+        if not isinstance(kubeconfig_val, str) or not kubeconfig_val.strip():
+            raise ConfigError(
+                "OpenTofu output 'kubeconfig' must be a non-empty string, got "
+                f"{type(kubeconfig_val).__name__}."
+            )
+        kubeconfig_yaml = kubeconfig_val
+
+        target_path = variables.get("kubeconfig_path")
+        if not target_path:
```
itssimrank: nit: this fallback can be extracted to a sub func and dedup everywhere else.

devops_bench/providers/vcluster.py
```diff
@@ -0,0 +1,417 @@
...
+            cluster_info: The cluster info of the cluster that was destroyed.
+            variables: Optional OpenTofu input variables used during provisioning.
+            success: Whether the stack destroy completed successfully.
+        """
+        vars_dict = variables or {}
+        host_kubeconfig = vars_dict.get("host_kubeconfig_path") or os.environ.get(
+            "HOST_KUBECONFIG", "~/.kube/config"
+        )
+        host_kubeconfig_path = str(Path(host_kubeconfig).expanduser().resolve())
+        host_context = vars_dict.get("host_kubecontext") or os.environ.get("HOST_KUBECONTEXT")
```
itssimrank: nit: use the repo's core get_env helper to keep consistency.

devops_bench/providers/vcluster.py
```diff
@@ -0,0 +1,417 @@
...
+                host_context = _get_current_context(host_kubeconfig_path)
+            except ConfigError:
+                host_context = None
+
+        if not success:
+            _log.info("Skipping VCluster PV deletion: destroy command failed.")
+        elif not cluster_info.name or not cluster_info.name.strip():
+            _log.warning("Skipping VCluster PV cleanup: cluster_info.name is empty.")
+        else:
+            label_selector = (
```
itssimrank: Should we change cleanup() to find PVs by ownership instead of labels — e.g., select PVs whose claimRef.namespace == "vcluster-<cluster_name>"?

itssimrank (approved):

devops_bench/providers/vcluster.py
isadominguez314: We stuck with labels mainly because k8s doesn't support field selectors on `claimRef.namespace` for PVs, so we'd have to list every PV in the whole cluster and filter client-side. 

Also, since PVs are cluster-scoped, using the explicit `devops-bench/*` labels (which we already sync via `values.yaml.tftpl`) is safer against accidental deletions and doesn't break if someone overrides the host namespace with a custom name.

tf/prebuilt/opa-remediation/variables.tf
```diff
@@ -61,3 +61,33 @@ variable "repo_path" {
...
+
+variable "service_type" {
+  type        = string
+  description = "Exposure mechanism for the vcluster (see modules/cluster). The harness's vcluster provider resolves this per host: NodePort for local hosts, LoadBalancer for remote ones. Must be declared here so that resolved value survives; undeclared variables are dropped before reaching tofu."
+  default     = "LoadBalancer"
+}
+
+variable "vcluster_service_cidr" {
+  type        = string
+  description = "Host cluster's Service CIDR for vcluster runs (see modules/cluster). The harness's vcluster provider auto-detects and injects this; empty keeps the chart default, which only suits kind-style hosts."
```
geojaz: The latest push removed the host service-CIDR detection from the Python provider (the probe, the VCLUSTER_SERVICE_CIDR env handling, and the injection in resolve_variables are all gone), but the Terraform side still carries the full plumbing: the vcluster_service_cidr variable, the serviceCIDR conditional in values.yaml.tftpl, and this description still says the harness auto-detects and injects it. As of this revision nothing populates the variable except a manual task variable.

That matters because the description documents the failure mode precisely: on a host whose Service CIDR differs from the Kubernetes default (GKE), the syncer blocks on "waiting for DNS service IP" and no pod ever syncs. I hit exactly this on a GKE Autopilot host on 2026-08-19, so with detection gone the provider only works out of the box on default-range hosts like kind.

Was the removal intentional (scope cut to local hosts for now) or lost in a rebase? If not, happy to help get the detection back in. If intentional, I'd keep the empirical documentation (it's accurate and hard-won) and just correct the auto-detection claim, e.g.:

```suggestion
  description = "Host cluster's Service CIDR for vcluster runs (see modules/cluster). Must be set explicitly on hosts that don't use the Kubernetes default range (e.g. GKE); empty keeps the chart default, which only suits kind-style hosts."
```

Plus a loud error or doc note for GKE hosts, so the next person doesn't get the very confusing silent syncer hang.

tf/modules/cluster/vcluster/variables.tf
```diff
@@ -0,0 +1,108 @@
...
+variable "chart_name_or_path" {
+  type        = string
+  description = "Helm chart name or local tarball path"
+  default     = "vcluster"
+}
+
+variable "chart_version" {
+  type        = string
+  description = "Helm chart version for vcluster"
+  default     = "0.20.0"
```
geojaz: chart_version defaults to 0.20.0, which is over a year old. Worth double-checking that experimental.syncSettings.syncLabels and the top-level serviceCIDR key are honored by that exact version: Helm ignores unknown values keys silently, so a schema mismatch fails at runtime, not at install. If there's no reason for 0.20.0 specifically, I'd suggest pinning a current 0.3x release; I ran 0.36.1 successfully.

tf/modules/cluster/vcluster/values.yaml.tftpl
```diff
@@ -0,0 +1,88 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+controlPlane:
+  service:
+    enabled: false
+  statefulSet:
```
geojaz: Heads-up from my #115 spike: on a GKE Autopilot host, the vcluster control plane statefulset needs explicit resource requests including ephemeral-storage, or Autopilot admission mutates/rejects it. The values here don't set a resources block for the control plane. Not a blocker if Autopilot hosts are out of scope for now, but worth a comment in the template either way so the next person doesn't rediscover it.

tf/prebuilt/opa-remediation/main.tf
```diff
@@ -46,22 +46,27 @@ locals {
...
+  vcluster_service_cidr = var.vcluster_service_cidr
+  node_port             = var.node_port
 }
 
 # Outside-the-cluster setup: install Kyverno, apply audit policies, deploy the
 # violating workloads, and seed the GitOps repo. Runs during `tofu apply`,
 # before the agent starts.
 resource "null_resource" "setup" {
-  depends_on = [module.cluster]
+  depends_on = [module.cluster, local_sensitive_file.vcluster_kubeconfig]
```
geojaz: This writes the raw kubeconfig to ${path.module}/.kubeconfig-vcluster-<name>, i.e. next to the stack sources. If the stack ever runs from the repo checkout rather than an isolated copy, that's a live credential sitting in the tree one `git add -A` away from a commit. Suggest writing it under TF_DATA_DIR (already per-run) or the system temp dir instead.

devops_bench/providers/vcluster.py
```diff
@@ -0,0 +1,410 @@
...
+                f"devops-bench/run-scoped=true,devops-bench/cluster-name={cluster_info.name}"
```
geojaz: +1 to the earlier syncLabels ordering comment, and I think the consequence is worth spelling out: nothing in this PR ever applies devops-bench/cluster-name=<name> to any object inside the vcluster, and syncLabels only copies labels that already exist on vcluster objects. Host PVs created by the provisioner from synced PVCs won't carry either label. So as written, this PV pass selects on labels that never exist and is a silent no-op. Since the module also disables the control plane's own PVC (persistence.volumeClaim.enabled: false), it might be simpler to drop the PV cleanup entirely until there's a labeling mechanism that provably tags them, so we don't carry dead safety code that looks load-bearing.

tf/prebuilt/opa-remediation/variables.tf
geojaz: The latest push removed the host service-CIDR detection from the Python provider (the probe, the VCLUSTER_SERVICE_CIDR env handling, and the injection in resolve_variables are all gone), but the Terraform side still carries the full plumbing: the vcluster_service_cidr variable, the serviceCIDR conditional in values.yaml.tftpl, and this description still says the harness auto-detects and injects it. As of this revision nothing populates the variable except a manual task variable.

That matters because the description documents the failure mode precisely: on a host whose Service CIDR differs from the Kubernetes default (GKE), the syncer blocks on "waiting for DNS service IP" and no pod ever syncs. I hit exactly this on a GKE Autopilot host on 2026-08-19, so with detection gone the provider only works out of the box on default-range hosts like kind.

Was the removal intentional (scope cut to local hosts for now) or lost in a rebase? If not, happy to help get the detection back in. If intentional, I'd keep the empirical documentation (it's accurate and hard-won) and just correct the auto-detection claim, e.g.:

```suggestion
  description = "Host cluster's Service CIDR for vcluster runs (see modules/cluster). Must be set explicitly on hosts that don't use the Kubernetes default range (e.g. GKE); empty keeps the chart default, which only suits kind-style hosts."
```

Plus a loud error or doc note for GKE hosts, so the next person doesn't get the very confusing silent syncer hang.

tf/modules/cluster/main.tf
```diff
@@ -44,3 +44,28 @@ module "kind" {
   node_count      = var.node_count
 }
 
+
+provider "kubernetes" {
```
geojaz: These provider "kubernetes" / "helm" blocks live inside tf/modules/cluster, which is itself a child module (every prebuilt stack calls it as module "cluster"). Terraform/OpenTofu treats provider configurations in non-root modules as legacy: the module can no longer be called with count/for_each, and removing or refactoring it later gets awkward because the provider config is destroyed with it. It also means every gcp/kind run now instantiates these providers pointed at pathexpand("~/.kube/config") even though nothing uses them.

Suggestion: declare the kubernetes/helm providers in the root stacks (tf/prebuilt/*) and pass them down with a providers map, keeping required_providers in the module. That keeps the module reusable and the host-cluster wiring visible at the stack level where credentials actually live.

tf/modules/cluster/vcluster/variables.tf
geojaz: chart_version defaults to 0.20.0, which is over a year old. Worth double-checking that experimental.syncSettings.syncLabels and the top-level serviceCIDR key are honored by that exact version: Helm ignores unknown values keys silently, so a schema mismatch fails at runtime, not at install. If there's no reason for 0.20.0 specifically, I'd suggest pinning a current 0.3x release; I ran 0.36.1 successfully.

devops_bench/providers/vcluster.py
```diff
@@ -0,0 +1,410 @@
...
+        is_local = _is_allowlisted_context(host_context, host_kubeconfig_path)
+        if not is_local:
+            if not get_bool("ALLOW_REMOTE_HOST_KUBECONTEXT", False):
+                raise ConfigError(
+                    f"Host kubecontext {host_context!r} is classified as remote. "
+                    "Set ALLOW_REMOTE_HOST_KUBECONTEXT=true to allow."
+                )
+            variables.setdefault("service_type", "LoadBalancer")
+        else:
+            variables.setdefault("service_type", "NodePort")
```
geojaz, line 401: Which cluster types has this been tested against to date? I ask because the NodePort default is the path a plain kind host takes, and a dynamically assigned NodePort on kind isn't reachable from the host unless the cluster was created with matching extraPortMappings, so I'd like to know whether that path has actually run.

Also, while this code seems to work, it might be helpful to be more descriptive about what's being built here: this is really about constructing the vcluster control plane endpoint and how that varies across different classes of clusters (locally reachable vs. needing a LoadBalancer), not about kind or GKE specifically. Would it be useful to name and structure it that way rather than per-provider (genuine question- and whether or not it would be- may not be appropriate for this PR)

tf/modules/cluster/vcluster/values.yaml.tftpl
geojaz, line 18: GKE Autopilot hosts shouldn't be out of scope here, and right now this template won't run on one: Autopilot admission requires explicit resource requests including ephemeral-storage on the control plane statefulset, and there's no resources block. I hit this on my #115 spike. Can we add one so Autopilot hosts work out of the box?

devops_bench/providers/vcluster.py
```diff
@@ -0,0 +1,410 @@
...
+                                return _is_local_server_url(server)
+        except Exception as exc:
+            _log.debug("Failed to classify kubecontext %s: %s", context_name, exc)
+
+    return context_name in _EXACT_LOCAL_CONTEXTS or any(
+        context_name.startswith(p) for p in _LOCAL_CONTEXT_PREFIXES
+    )
+
+
+def _default_kubeconfig_path(cluster_name: str) -> str:
```
geojaz, line 135: With BENCH_PARALLEL=true this returns $KUBECONFIG verbatim, and ensure_cluster_credentials then O_TRUNC-writes the vcluster kubeconfig to that path. The guard only refuses ~/.kube/config exactly, so a user whose KUBECONFIG points at a hand-maintained merged config anywhere else gets it silently overwritten (and cleanup() may later delete it if it happens to live under a scratch root).

Suggestion: always derive a per-run path (e.g. $TMPDIR/vcluster-<cluster>-kubeconfig.yaml) and let the parallel harness point KUBECONFIG at it, rather than writing into whatever KUBECONFIG already names. Alternatively refuse to overwrite any pre-existing file the harness didn't create.

devops_bench/providers/vcluster.py
```diff
@@ -0,0 +1,410 @@
...
+        return True
+    try:
+        ip = ipaddress.ip_address(hostname)
+        return ip.is_private or ip.is_loopback
+    except ValueError:
+        pass
+    return False
+
+
+def _is_explicit_public_ip(server: str) -> bool:
```
geojaz: I don't think this method actually ends up being used. Is it still required?

tf/modules/cluster/vcluster/main.tf
```diff
@@ -0,0 +1,166 @@
...
+
+resource "kubernetes_resource_quota" "vcluster_quota" {
+  metadata {
+    name      = "vcluster-quota"
+    namespace = kubernetes_namespace.vcluster.metadata[0].name
+  }
+
+  spec {
+    hard = {
+      "limits.cpu"             = "7"
```
geojaz: The bench's convention for sizing elsewhere is tofu variables that tasks can override (node_count / machine_type via infrastructure.variables in the task YAML); these quota and LimitRange numbers are the vcluster equivalent of that, but hardcoded in the module. Could we expose them the same way, with these values as defaults, so a bigger host or a heavier task seed doesn't need a module edit?

Related: we already have quantity-parsing art in the bench (to_number() in the verification module), and I opened #116 to promote it to a shared utility. That would pair well with making these values variables if we ever want to validate or derive them harness-side.

tf/modules/cluster/vcluster/variables.tf
```diff
@@ -0,0 +1,108 @@
...
+}
+
+variable "host_kubecontext" {
+  type        = string
+  description = "Host Kubernetes context to use"
+  default     = null
+  nullable    = true
+}
+
+variable "infra_provider" {
```
geojaz: infra_provider, project_id, and kubeconfig_path are declared in this module but never referenced in it. Suggest dropping them; the harness-facing declarations already live at the stack level.

tf/prebuilt/opa-remediation/main.tf
geojaz, line 89: This writes the raw kubeconfig next to the stack sources; if the stack ever runs from the repo checkout rather than an isolated copy, that's a live credential one `git add -A` away from a commit. Suggest writing it under TF_DATA_DIR (already per-run) or the system temp dir instead. Longer term this feels like something the sandbox work in #72 should own: run-scoped credentials that never touch the tree.

devops_bench/providers/vcluster.py
geojaz: Could we identify dangling PVs via their claimRef instead of labels? PVs are cluster-scoped, so they won't inherit labels from anything in the vcluster, and I don't see anything applying devops-bench/cluster-name inside the vcluster (syncLabels only copies labels that already exist there), so I suspect this selector never matches and the cleanup silently no-ops. Since the vcluster namespace is tofu-managed and deleted on destroy, PVs whose spec.claimRef.namespace points at that namespace are exactly the orphan set. That's not label-selectable, but a jsonpath filter over `kubectl get pv -o json` gets there.

geojaz (commented):
Looks good- left a few questions and suggestions that should be relatively straightforward...

devops_bench/providers/vcluster.py
isadominguez314: To date, I've tested `opa-remediation` end-to-end on vCluster with a GKE host (`LoadBalancer` path) and separately on a standalone KinD cluster. I haven't run vCluster on a local KinD host with `NodePort` yet.I  agree that framing endpoint resolution around reachability (locally reachable vs. remote requiring LoadBalancer/ingress) is a much better abstraction than something provider specific. Happy to update this later once the core functionality is in place

devops_bench/providers/vcluster.py
```diff
@@ -0,0 +1,411 @@
...
+            return False
+
+        tmp_dir = Path(tempfile.gettempdir()).resolve()
+        bench_root = get_env("BENCH_RUN_STATE_ROOT")
+        tf_data = get_env("TF_DATA_DIR")
+
+        return (
+            resolved.parent == tmp_dir
+            or (tmp_dir / "devops-bench-runs") in resolved.parents
+            or (bool(bench_root) and Path(bench_root).resolve() in resolved.parents)
```
janetkuo, line 265: In `_is_safe_scratch_path`, `resolved.parent == tmp_dir` unconditionally marks any file directly under `/tmp` as scratch. Restrict to `("vcluster-", "kubeconfig")` prefixes (matching line 271) to avoid deleting custom user-specified kubeconfig paths in `/tmp`.

devops_bench/providers/vcluster.py
```diff
@@ -0,0 +1,411 @@
...
+                ) from exc
+            f.write(kubeconfig_yaml)
+        _log.info("Wrote virtual cluster kubeconfig to %s (mode 0600)", resolved_target)
+        os.environ["KUBECONFIG"] = str(resolved_target)
+
+        return ClusterInfo.from_dict(
+            {
+                "name": cluster_name,
+                "location": location,
+                "project": "local-vcluster",
```
janetkuo: Respect `variables.get("project_id")` so remote GKE host contexts retain their configured project ID in `ClusterInfo`:

```suggestion
                "project": variables.get("project_id") or "local-vcluster",
```

devops_bench/providers/vcluster.py
```diff
@@ -0,0 +1,411 @@
...
+    if not server:
+        return False
+    try:
+        split_res = urlsplit(server)
+        hostname = split_res.hostname
+    except Exception:
+        return False
+    if not hostname:
+        return False
+    if hostname == "localhost":
```
janetkuo, line 75: If the server URL uses a local hostname like `kubernetes.docker.internal` (standard for Docker Desktop on macOS/Windows), `ipaddress.ip_address` raises `ValueError` and returns `False`. Checking for these standard hostnames ensures they are recognized as local without needing `ALLOW_REMOTE_HOST_KUBECONTEXT=true`

janetkuo (approved):
/lgtm

devops_bench/providers/vcluster.py
isadominguez314: Addressed in PR #130

PR #79, opened 2026-08-07
fix(verification): pin checks to the run's cluster and stop scoring a broken pass

geojaz:
Two related correctness problems in how a run's checks reach the cluster and how a broken run gets scored.

Checks were resolving their cluster through ambient kubeconfig context rather than the cluster the run actually provisioned. On a workstation with more than one cluster in kubeconfig, or when two runs overlap, a check could observe the wrong cluster and report a confident result about something it never touched. This threads the run's cluster identity through the context and the provider layer so every check is pinned to the cluster the run owns.

Separately, a run that failed to produce a usable result could still reach the judge and be scored as a pass. That turns an infrastructure failure into a silent correctness signal, which is the worst direction for the error to go. A broken run now stops before scoring instead.

Touches the context and provider plumbing, the verification base, and the `pod_healthy`, `resource_property`, and `scaling_complete` verifiers. Unit tests cover the pinning behavior and the withheld-score path.

### DIFF
--- a/devops_bench/core/context.py
+++ b/devops_bench/core/context.py
@@ -51,12 +51,19 @@ class ClusterInfo:
         project: Cloud project; None for local clusters.
         kubeconfig_path: Kubeconfig path; resolved from ``KUBECONFIG`` or
             ``~/.kube/config`` when not supplied.
+        context: Kubeconfig context naming this specific cluster (e.g.
+            ``kind-<cluster>`` or ``gke_<project>_<location>_<cluster>``).
+            A kubeconfig file alone does not pin a cluster: it can carry
+            several contexts, or none selected as current. ``None`` when the
+            provider could not resolve one, in which case callers fall back
+            to the ambient current-context.
     """
 
     name: str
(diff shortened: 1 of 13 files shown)


geojaz:
Noting a follow-up that is deliberately not in this PR, so the gap does not read as an oversight.

This pins every verifier that exists on main today. Two more, `http_probe` and `external_http_probe`, are still in flight in #55 and #54, and they are not pinned here because their source files do not exist on this branch yet.

The commit that pins them is written and tested. It needs #54, #55 and this PR all present at the same time, since it patches both verifier files and reads the `context` field this PR adds to `BaseVerifier`. No single branch has that combination today, so it is parked until they merge. I will open it as a small follow-up PR once the last of the three lands.

I did try splitting it by file onto #54 and #55 so nothing would be left waiting. That does not work: neither of those branches has the `context` field, so each half fails its own tests. The dependency is real rather than cosmetic, which is why it is sequenced instead of forced.

PR #84, opened 2026-08-07
verification: define mode hold for both objective and safeguard roles

geojaz:
## What this does

Implements `mode: hold` for verification entries. The original version of this PR only handled safeguards; it now covers both roles, because working through #59 made clear they are two different things wearing one keyword.

**`safeguard` + `hold`** is an invariant that must never break. It is sampled on a background thread across the agent's turn, so a violation that happens and then self-corrects is still caught. `violated` is sticky. There is no window to configure: the window is the turn.

**`objective` + `hold`** is a thing that starts false, must become true, and must then stay true. It is soaked after the agent's turn, for an explicitly declared `hold_window_sec`, drawing from the shared post-run verification budget.

Routing an objective through the live monitor is a bug, not a variation: the first sample fails before the agent has done anything and latches a permanent violation. Before this change that combination validated cleanly and would have failed every objective it was applied to.

## Structure

`safeguard_monitor.py` becomes `hold.py`, since the mechanism was never safeguard-specific (it filtered on mode, never on role). Both drivers share one `_fold_sample` and one `hold_verdict`, so the outcome rule is defined exactly once.

## Verdict rule

Checked in order: zero samples is an error; all samples errored is an error; a violation is a fail; a window that ends on an unevaluable sample is an error; otherwise a pass.

The ordering is deliberate. An error that recovers within the window is observation noise, typically a transient kubectl failure, and is absorbed. An error that never clears means the entry was not actually observed when the window closed, which is not a pass. But a confirmed violation outranks both: losing observability afterward does not un-observe it. That ordering matters downstream, where an objective reporting `error` nulls a task's correctness entirely while `fail` scores as a fail.

Previously an entry with 49 errored samples and 1 clean pass scored identically to 50 clean passes. It no longer does.

## Breaking change to spec validation

Two combinations that used to parse are now rejected:

- `role: objective` with `mode: hold` requires `hold_window_sec`. There is no default on purpose: a silent default would quietly consume the shared verification budget on every task in a suite.
- `role: safeguard` with `mode: hold` must not set `hold_window_sec`, whose window is always the agent's turn. Accepting and ignoring it would mislead.

Any existing spec using either combination will now fail to parse, with an error message explaining why.

## Replaces #59

#59 was my own earlier, independent implementation of `mode: hold` as a post-run sampling loop. Both PRs branched from the same commit and both removed the same "not yet supported" guard, so they could not both land. #59's post-run soak survives here as `run_hold_window`, serving objective holds. I am closing #59 in favor of this one.

## Summary by CodeRabbit

* **New Features**
  * Added hold-mode verification for safeguards and objectives.
  * Safeguards are monitored during execution, while objectives are observed for a configurable period afterward.
  * Added configurable polling intervals and hold windows.
  * Verification results now include violation details, sample counts, timing, and evaluation errors.

* **Bug Fixes**
  * Improved handling of transient violations, repeated evaluation errors, agent failures, deadlines, and monitor cleanup.
  * Hold configuration now validates timing values and required settings.
  * Hold results consistently distinguish violations from evaluation errors.

### DIFF
--- a/devops_bench/evalharness/default.py
+++ b/devops_bench/evalharness/default.py
@@ -48,6 +48,7 @@ from devops_bench.deployers.factory import get_deployer
 from devops_bench.evalharness.artifacts import collect_generated_files, snapshot_dir
 from devops_bench.evalharness.base import Harness
 from devops_bench.evalharness.reporter import ResultReporter
+from devops_bench.evalharness.safeguard_monitor import HoldObservation, SafeguardMonitor
 from devops_bench.evalharness.scenario import (
     VERIFICATION_TIMEOUT_SEC,
     VERIFICATION_TOTAL_BUDGET_SEC,
@@ -439,6 +440,8 @@ class DefaultEvalHarness(Harness):
         self,
         entries: list[VerificationEntry],
         timeout_sec: float = VERIFICATION_TIMEOUT_SEC,
+        *,
+        hold_observations: dict[str, HoldObservation] | None = None,
     ) -> list[dict[str, Any]]:
         """Evaluate every entry against the live cluster after the agent finishes.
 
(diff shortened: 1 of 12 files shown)


geojaz:
@coderabbitai full review

PR #87, opened 2026-08-11
feat(agents): add the Hermes CLI agent harness

eugeneng04:
## What

Adds a `hermes` CLI agent harness, ported from gke-labs/devops-bench#178 and rewired onto this repo's current interfaces.

Hermes (NousResearch/hermes-agent) is driven as a subprocess; everything the run needs is laid down in a run-scoped `$HERMES_HOME` so a benchmark run never reads or writes the developer's `~/.hermes`.

## How capabilities are wired

| Capability | Channel |
|---|---|
| MCP servers | `mcp_servers` entries in `$HERMES_HOME/config.yaml` (plus `KUBECONFIG` passed through to each server's env) |
| Skills | materialized to `$HERMES_HOME/skills/<name>/SKILL.md` |
| Rules | prepended to the prompt — `hermes chat` has no system-prompt flag |
| Model auth | `config.api_key` exported to the env vars the resolved `ProviderSpec` names |

`HERMES_HOME` is `<workspace>/.hermes`, so hermes still runs with `cwd` set to the harness workspace (relative paths in a task resolve where the harness expects) while its config and session DB stay in a single subdirectory instead of scattering across the artifact diff.

## Token buckets

Trajectory and usage are read back from the run's SQLite `state.db`. The `sessions` columns map onto the canonical buckets from `devops_bench.agents.result`:

| `sessions` column | bucket |
|---|---|
| `input_tokens` | `input` |
| `cache_read_tokens` | `cached` |
| `cache_write_tokens` | `cache_write` |
| `reasoning_tokens` | `reasoning` |
| `output_tokens` | `output` |
| — | `total` (sum of the reported buckets) |

Counts are summed across all session rows, since the DB is run-scoped and a run may write more than one. Unreported buckets stay `None` rather than becoming a fabricated `0` — a missing DB, an older schema without the token columns, or any read failure all yield all-`None`.

`extract_tokens_from_db` opens the DB read-only (`mode=ro` URI) so a live `state.db` is never locked.

## Testing

46 unit tests in `tests/unit/agents/test_agents_cli_hermes.py` covering registration, provider/model argv mapping, config generation and merge, skills/MCP materialization, the four `_execute` exit paths (success, non-zero exit, timeout, missing binary), trajectory pairing and its malformed-input paths, and every token bucket. Two guards worth calling out:

- `test_tokens_fill_every_canonical_bucket` fails if `TOKEN_BUCKETS` gains a bucket this parser skips.
- `test_tokens_reach_the_result_row_unchanged` runs the parser output through `normalize_tokens`, so a renamed bucket fails loudly here instead of silently reading `None` on the dashboard row.

Full suite: 1227 passed. `ruff check` and `ruff format` clean.

## Notes

- The token mapping assumes hermes normalizes provider usage to the canonical split (`input` excluding cache reads, `output` excluding reasoning — the Anthropic convention its column names mirror). Hermes is multi-provider; if it instead stores provider-native usage on some backends, `total` would double-count there. Flagged in a comment in `parsing.py`; unverified against a live run.
- The upstream PR's docs changes are dropped — neither target file exists in this repo.

## Summary by CodeRabbit

* **New Features**
  * Added support for running evaluations with the Hermes CLI agent.
  * Added provider-aware configuration and validation for Hermes.
  * Added isolated execution with configurable credentials, MCP servers, skills, and prompt rules.
  * Added token usage and execution trajectory reporting.
  * Registered Hermes in the default agent selection.

* **Bug Fixes**
  * Improved MCP environment isolation for CLI agents, including cloud configuration settings.
  * Ensured Hermes runs use fresh evaluation configuration and reliable provider validation.

* **Tests**
  * Expanded coverage for execution, configuration, error handling, token usage, and trajectory parsing.

### DIFF
--- a/devops_bench/agents/cli/hermes/__init__.py
+++ b/devops_bench/agents/cli/hermes/__init__.py
@@ -1,0 +1,26 @@
+# Copyright 2026 The Kubernetes Authors.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""Hermes CLI agent package.
+
+Importing this package self-registers the agent under the ``"hermes"`` key.
+"""
+
+from devops_bench.agents.cli.hermes.agent import HermesAgent
+from devops_bench.agents.cli.hermes.parsing import (
+    extract_tokens_from_db,
+    extract_trajectory_from_db,
+)
+
+__all__ = ["HermesAgent", "extract_tokens_from_db", "extract_trajectory_from_db"]
--- a/devops_bench/agents/cli/hermes/agent.py
+++ b/devops_bench/agents/cli/hermes/agent.py
@@ -216,6 +216,22 @@ class HermesAgent(AgentHarness):
         candidate = os.path.expanduser("~/.local/bin/hermes")
         return candidate if os.path.exists(candidate) else "hermes"
 
+    def _mcp_isolation_env(self) -> dict[str, str]:
+        """Resolve the isolation vars an MCP server needs, in the CLI's own order.
+
+        ``run_env`` publishes the per-run ``KUBECONFIG`` / ``CLOUDSDK_CONFIG`` on
+        ``os.environ``, while ``config.extra_env`` is the operator's override and
+        wins in :func:`_build_env`. Reading only one of the two would hand the
+        MCP servers a different cluster (or gcloud config) than hermes itself
+        got, so both are consulted in the same precedence.
+        """
+        resolved: dict[str, str] = {}
+        for key in _MCP_ISOLATION_ENVS:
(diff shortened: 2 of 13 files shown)


eugeneng04:
Thanks — addressed all six in 6b9daf5.

**`parsing.py` — reasoning double-count.** Real bug, and the most important one here. Confirmed against `NousResearch/hermes-agent@main:agent/usage_pricing.py`: `CanonicalUsage.total_tokens` is `prompt_tokens + output_tokens`, where `output_tokens` is the provider completion total and `reasoning_tokens` is a slice of it — reasoning is never added on top. Our canonical `output` bucket excludes reasoning, so the parser now derives `output = output_tokens - reasoning_tokens` (clamped at 0). `total` now equals Hermes's own figure. Also dropped the now-obsolete "unverified" caveat comment, and added a clamp test for a provider reporting reasoning > completion.

**`agent.py` — non-mapping seeded `config.yaml`.** Real; `or {}` only catches `None`. Guarded with an `isinstance(loaded, dict)` check plus a warning, with a test seeding `- a\n- b\n`.

**`agent.py` — state-isolation docstring.** Correct, the claim was wrong: the collector does include dot-directories. Reworded to say `.hermes` *is* collected, as a single directory rather than scattered across the diff — which is the actual reason it moved out of the workspace root.

**Nitpicks** — all three taken: `quote()` on the SQLite URI path (with a `run?id=1` regression test), `:g` on the timeout seconds, and type annotations on the test helpers and every `fake_run` stub.

1230 tests pass; `ruff check` and `ruff format` clean.

eugeneng04:
@coderabbitai review

PR #88, opened 2026-08-11
feat(agents): add the Claude Code CLI agent harness

eugeneng04:
Adds a `claude` CLI agent harness that drives the `claude` binary in headless
mode and folds its event stream into the canonical trajectory and token
buckets. Ported from gke-labs/devops-bench#199 and rewired onto the shared
token schema that landed here since.

## What it does

Runs `claude -p --output-format stream-json --verbose` and parses the wrapped
SDK event stream on stdout — no session-file reads off disk. Capabilities go
through Claude Code's native cwd channels, written into the per-run working
directory:

| Capability | Channel |
| --- | --- |
| Rules | `CLAUDE.md` in the cwd |
| Skills | `<cwd>/.claude/skills/<name>/SKILL.md` |
| MCP servers | `<cwd>/.claude/mcp-config.json` via `--mcp-config` |

Auth is env-driven through the shared provider contract: `config.api_key` onto
the provider's key env var(s), or keyless Vertex / Bedrock via ADC / AWS
credentials. `CLAUDE_CONFIG_DIR` is redirected to a fresh per-run temp dir so
the CLI's mutable global state never races across concurrent evals.

## Token buckets

The upstream PR carried a harness-local bucket tuple pending the shared schema.
This version maps the Anthropic `usage` block straight onto `TOKEN_BUCKETS`,
with cache reads and cache writes kept in separate buckets (writes bill at a
premium):

| Provider field | Bucket |
| --- | --- |
| `input_tokens` | `input` |
| `cache_read_input_tokens` | `cached` |
| `cache_creation_input_tokens` | `cache_write` |
| `output_tokens_details.thinking_tokens` | `reasoning` |
| `output_tokens` less the above | `output` |

Extended thinking is billed *inside* `output_tokens` and counted again under
`output_tokens_details`, so it is subtracted back out — `output` excludes
`reasoning` per the contract, and `total` still equals the provider's own
accounting. An unreported bucket stays `None` rather than a fabricated `0`. A
cross-layer test asserts the emitted keys survive `normalize_tokens` onto the
dashboard row.

## Alias canonicalization

`claude-code` joins `gemini-cli` as a friendly alias. Both the registry lookup
and the manifest write now go through one `_canonical_agent_type` helper, so an
arm selected by alias aggregates under the same `harness` / `setup_id` as the
canonical key instead of splitting into a second dashboard setup.

## Review fixes

A review panel over the port turned up seven issues, each reproduced against
the real `claude` binary (2.1.228) before fixing, and fixed in the second
commit:

- **Thinking tokens misfiled into `output`.** The port assumed Anthropic reports
  no separate thinking count. A live run returns
  `"output_tokens":3069,"output_tokens_details":{"thinking_tokens":2778}`, and
  the thinking count is included in `output_tokens`.
- **The per-turn usage accumulator reported `output` ~1000x too low.** Per-turn
  `usage.output_tokens` on an assistant envelope is the `message_start`
  placeholder — a summed 3 against a terminal 3069. It only fires on the
  timed-out path, so every timed-out run recorded a plausible-looking wrong
  number. The bucket is now left unreported; the prompt-side fields do
  accumulate faithfully and are kept.
- **`str.splitlines` shredded events containing U+0085.** The CLI escapes U+2028
  but leaves NEL raw inside JSON strings, which real command output carries when
  a log is mis-decoded as latin-1. On a reproduced stream that cost 2 of 3 tool
  calls and injected 6 bogus errors — enough to flip `validated` to `False` and
  drop the run off the leaderboard. Framing now keys on `\n` alone.
- **Tool results retained verbatim.** Claude Code echoes the full body of every
  Read/Bash result (measured ~20 KB per call, one Read at 58 KB), and the whole
  trace is `json.dumps`'d into two LLM judge prompts with nothing truncating in
  between. Clipped to a head+tail slice with the middle marked elided.
- **`--strict-mcp-config` skipped on the baseline arm.** It was only passed
  alongside `--mcp-config`. Verified: without it a stray `.mcp.json` in the
  workspace loads; with it, `mcp_servers` is empty. That silently granted MCP
  tools to the un-augmented arm, contaminating the comparison the bench exists
  to measure. Now unconditional.
- **No `--` before the prompt.** A prompt whose first token starts with `-` was
  read as a flag and aborted the run with rc=1 and no output.
- **Uncapped child stderr persisted to results.** `metadata["stderr"]` was
  clipped at 2000 chars while the same stderr went unbounded into `errors`,
  which is what reaches disk. Both paths now share one cap.

Also verified and deliberately left alone: `CLOUD_ML_REGION` is the correct
Vertex region variable for this CLI (`ANTHROPIC_VERTEX_LOCATION` does not
appear in the binary at all).

## Testing

- 66 unit tests for this harness; 1248 across the suite, all passing.
- `ruff check` / `ruff format` clean.
- End-to-end against the real binary in a workspace seeded with a stray
  `.mcp.json`: clean run, one matched tool call, stray server suppressed, and
  `total` equal to the sum of the reported buckets.

## Summary by CodeRabbit

* **New Features**
  * Added support for running evaluations with the Claude Code CLI.
  * Added structured capture of assistant responses, tool activity, terminal output, token usage, and errors.
  * Added isolated per-run configuration and MCP setup for more reliable execution.
  * Added support for resolving the `claude-code` alias to the canonical Claude harness.

* **Documentation**
  * Updated provider configuration and turn-limit guidance for Claude CLI usage.

* **Bug Fixes**
  * Improved handling of partial output, failed commands, and token reporting.

* **Tests**
  * Added comprehensive coverage for parsing, execution, error handling, configuration, and registry resolution.

### DIFF
--- a/devops_bench/agents/cli/claude_code/agent.py
+++ b/devops_bench/agents/cli/claude_code/agent.py
@@ -30,7 +30,8 @@ per-run working directory before invocation:
   document at ``<cwd>/.claude/mcp-config.json``, passed via ``--mcp-config``.
   ``--strict-mcp-config`` is always set — including on a baseline arm with no
   bindings — so any stray project ``.mcp.json`` is ignored and no trust prompt
-  fires.
+  fires. A bound run first checks the binary against
+  :data:`_MCP_WAIT_MIN_VERSION`.
 
 Auth is env-driven, matching the bench contract: ``config.api_key`` →
 ``ANTHROPIC_API_KEY`` for the direct API, or keyless Vertex / Bedrock via ADC /
@@ -44,6 +45,7 @@ from __future__ import annotations
 import contextlib
 import json
 import os
+import re
 import tempfile
 from collections.abc import Iterator
 from pathlib import Path
@@ -83,6 +85,32 @@ _STDERR_TAIL_CHARS = 2000
 def _stderr_tail(stderr: str | None) -> str:
     """Stripped last :data:`_STDERR_TAIL_CHARS` characters of ``stderr``."""
     return (stderr or "").strip()[-_STDERR_TAIL_CHARS:]
+
+
+# ``--mcp-config`` under ``-p`` only waits for still-pending servers to connect
+# before the first turn from this version on. An older binary accepts the flag
+# and starts the turn anyway, so an MCP-augmented arm can run with none of its
+# tools attached and still be scored as augmented — the same silent
+# contamination ``--strict-mcp-config`` prevents from the other direction.
+_MCP_WAIT_MIN_VERSION = (2, 1, 221)
+_VERSION_RE = re.compile(r"(\d+)\.(\d+)\.(\d+)")
+
+
+def _claude_version(target: str) -> tuple[int, int, int] | None:
+    """Parse ``claude --version``, or ``None`` when it cannot be determined.
+
+    An unreadable version is not an error: ``config.target`` may be a wrapper
+    script with its own ``--version`` surface, and refusing to run on a probe
+    that merely failed to parse would be worse than the risk it guards.
+    """
+    try:
+        completed = run([target, "--version"], check=False, timeout=30)
+    except (OSError, SubprocessError):
+        return None
+    match = _VERSION_RE.search(completed.stdout or "")
+    if completed.returncode != 0 or not match:
+        return None
+    return (int(match[1]), int(match[2]), int(match[3]))
 
 
 def _errored_with_tokens(msg: str, *, stderr: str | None = None) -> AgentResult:
@@ -280,6 +308,16 @@ class ClaudeCodeAgent(AgentHarness):
                 # not leave it at the umask default (0o644 on most machines).
                 mcp_path.chmod(0o600)
                 mcp_config_path = str(mcp_path)
+                version = _claude_version(target)
+                if version is not None and version < _MCP_WAIT_MIN_VERSION:
+                    return _errored_with_tokens(
+                        "claude "
+                        + ".".join(str(part) for part in version)
+                        + " predates the --mcp-config startup wait (needs "
+                        + ".".join(str(part) for part in _MCP_WAIT_MIN_VERSION)
+                        + "); an MCP-bound arm would run without its servers. "
+                        "Upgrade the binary or drop the mcp_servers binding."
+                    )
 
             argv = _build_argv(
                 target,
--- a/devops_bench/agents/cli/claude_code/agent.py
+++ b/devops_bench/agents/cli/claude_code/agent.py
@@ -83,6 +83,14 @@ _log = get_logger("agents.cli.claude_code")
 # ``metadata`` and ``errors``), so every path clips it to the same tail.
 _STDERR_TAIL_CHARS = 2000
 
+# Under ``-p`` the CLI reads a piped prompt from stdin and waits out a 3s
+# timeout before giving up when stdin stays open with nothing on it — which is
+# what an inherited stdin looks like. Handing it an empty string closes the pipe
+# immediately: ~3.7s saved per task, and no "no stdin data received" warning
+# polluting ``metadata["stderr"]`` (and, on a non-zero exit, the error message,
+# where it would displace the real cause).
+_CLOSED_STDIN = ""
+
 
 def _stderr_tail(stderr: str | None) -> str:
     """Stripped last :data:`_STDERR_TAIL_CHARS` characters of ``stderr``."""
@@ -348,6 +356,7 @@ class ClaudeCodeAgent(AgentHarness):
                         cwd=workdir,
                         check=False,
                         timeout=self.config.timeout_sec,
+                        input=_CLOSED_STDIN,
                     )
                 except SubprocessError as exc:
                     # str(exc) embeds the child's full stderr, so rebuild the
--- a/devops_bench/agents/cli/claude_code/agent.py
+++ b/devops_bench/agents/cli/claude_code/agent.py
@@ -36,8 +36,7 @@ Auth is env-driven, matching the bench contract: ``config.api_key`` →
 ``ANTHROPIC_API_KEY`` for the direct API, or keyless Vertex / Bedrock via ADC /
 AWS credentials. ``CLAUDE_CONFIG_DIR`` is redirected to a fresh per-run temp dir
 so Claude Code's mutable global state never races across concurrent evals (see
-:func:`_claude_config_dir` for the OAuth-debug escape hatch, and
-``docs/appendix/known_issues.md`` for the keyless-Vertex ``--parallel`` gotcha).
+:func:`_claude_config_dir` for the OAuth-debug escape hatch).
 """
 
 from __future__ import annotations
@@ -276,6 +275,10 @@ class ClaudeCodeAgent(AgentHarness):
                 claude_dir.mkdir(parents=True, exist_ok=True)
                 mcp_path = claude_dir / _CLAUDE_MCP_FILE
                 mcp_path.write_text(json.dumps({"mcpServers": servers}, indent=2), encoding="utf-8")
+                # A binding's argv can carry a server credential, and this file
+                # lands in the run workspace the harness later collects, so do
+                # not leave it at the umask default (0o644 on most machines).
+                mcp_path.chmod(0o600)
                 mcp_config_path = str(mcp_path)
 
             argv = _build_argv(
--- a/devops_bench/agents/cli/claude_code/agent.py
+++ b/devops_bench/agents/cli/claude_code/agent.py
@@ -37,7 +37,8 @@ Auth is env-driven, matching the bench contract: ``config.api_key`` →
 ``ANTHROPIC_API_KEY`` for the direct API, or keyless Vertex / Bedrock via ADC /
 AWS credentials. ``CLAUDE_CONFIG_DIR`` is redirected to a fresh per-run temp dir
 so Claude Code's mutable global state never races across concurrent evals (see
-:func:`_claude_config_dir` for the OAuth-debug escape hatch).
+:func:`_claude_config_dir` for the OAuth-debug escape hatch, which ``--parallel``
+overrides).
 """
 
 from __future__ import annotations
@@ -60,6 +61,7 @@ from devops_bench.agents.shared.cli_capabilities import (
     materialize_skills,
 )
 from devops_bench.core import SubprocessError, get_logger
+from devops_bench.core.config import get_bool
 from devops_bench.core.model_providers import resolve_provider
 from devops_bench.core.subprocess import run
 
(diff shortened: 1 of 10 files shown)


eugeneng04:
Thanks — addressed in 8499755. Four of the five actionable items applied, one skipped with reasoning below.

**Applied**

- Dropped the dangling `docs/appendix/known_issues.md` pointer from the module docstring — that file doesn't exist in this tree; it was lost in the port.
- `chmod 0o600` on the written `.claude/mcp-config.json`. A binding's argv can carry a server credential and the file lands in the run workspace the harness later collects, so the umask default (0o644 on most machines) was wrong. Asserted via `stat.S_IMODE` in the existing MCP test.
- `monkeypatch.delenv("CLOUD_ML_REGION", raising=False)` in the Vertex region-default test — an ambient value on a dev machine or CI runner would shadow the fallback the test is checking.
- Resolve-before-read in the alias tests. Confirmed both failed in isolation with `NotRegisteredError: 'claude' is not registered in the 'agents' registry`; builtins self-register on the lazy import that `resolve_agent` performs, so reading `AGENTS` first only passed when an earlier test happened to import the module. Fixed the pre-existing `gemini-cli` one too rather than leaving a known-broken sibling.
- Nit: annotated the `monkeypatch` / `tmp_path` fixture params (repo convention is 121 annotated vs 34 bare).

**Not applied**

Gating the run on Claude Code v2.1.221 for `--mcp-config`. The CLI reference attaches that floor to the wait-for-pending-servers behaviour — "The wait requires Claude Code v2.1.221 or later" — not to the flag itself, so a version probe would add a subprocess call per run for nothing.

`ruff check` / `ruff format` clean, 1248 tests passing, and both alias tests pass when run alone.

devops_bench/agents/cli/claude_code/agent.py
```diff
@@ -0,0 +1,344 @@
...
+        "--dangerously-skip-permissions",
+        "--strict-mcp-config",
+    ]
+    if model:
+        argv.extend(["--model", model])
+    if max_turns is not None and max_turns > 0:
+        argv.extend(["--max-turns", str(max_turns)])
+    if mcp_config_path:
+        argv.extend(["--mcp-config", mcp_config_path])
+    argv.extend(["--", prompt])
```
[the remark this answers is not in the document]
eugeneng04: Revisited the one item I'd skipped — the v2.1.221 floor — and implemented it in 704a078. My original reasoning was right on the doc reading but wrong on the impact.

The CLI reference does attach the floor to the wait, not to the flag:

> When you pass this flag with `-p`, Claude Code waits for still-pending servers to connect before running the first turn, up to the `MCP_TIMEOUT` startup timeout, 30 seconds by default […] **The wait requires Claude Code v2.1.221 or later**

So an older binary accepts `--mcp-config` and runs. But losing the wait is itself a correctness bug for this benchmark: the first turn starts before the servers connect, so an MCP-augmented arm can execute with none of its tools attached and still be scored as augmented. That's the same silent contamination `--strict-mcp-config` prevents from the other direction, and it's worse than a loud failure — a wrong number on the leaderboard beats no number only if it's right.

**What landed**

- `_claude_version()` parses `claude --version`; the run is refused below `(2, 1, 221)` with an actionable message naming both versions.
- The probe is gated on `mcp_config_path` being set, so a baseline arm pays no extra subprocess.
- An inconclusive probe — non-zero exit, unparseable output, or spawn failure — proceeds rather than blocking. `config.target` accepts any binary and may be a wrapper script with its own `--version` surface; refusing to run because a probe failed to yield a number would cost more than the risk it guards.

Verified against the real binary: `_claude_version('claude')` → `(2, 1, 228)`, no block; a missing binary → `None`, proceeds. Five new tests cover the refusal, the baseline skip, and all three inconclusive-probe paths.

`ruff check` / `ruff format` clean, 1253 tests passing.

eugeneng04:
Both outside-diff comments addressed in 253e0ef.

**Type annotations on the `fake_run` doubles** — done, all 19 in this file now carry `(argv: list[str], **kwargs: object) -> SimpleNamespace`.

Worth flagging for the repo rather than this PR: the convention across `tests/` is currently 66 bare nested doubles to 12 annotated, and the sibling `test_agents_cli_gemini.py` is entirely bare with this exact signature shape. AGENTS.md does say "all Python code must include type hints", so this file is now the compliant one — but the guideline and the codebase disagree, and it's worth settling that in one sweep instead of file by file as PRs touch them.

**`CLAUDE_CONFIG_DIR` isolation** — the hazard is real and I've fixed it, but not by always isolating.

Always generating a per-run dir would leave an OAuth-authenticated operator unable to run the bench at all without an API key: credentials live in the config dir that the fresh temp dir replaces. That's the entire reason the escape hatch exists, so removing it trades one breakage for another.

What actually makes the hatch dangerous is concurrency, and this bench's concurrency is explicit and opt-in — several benchmark processes on one host under `--parallel` / `BENCH_PARALLEL`, which is the exact collision class `core/run_env.py` already isolates `KUBECONFIG` / `CLOUDSDK_CONFIG` / `TF_DATA_DIR` against. So the hatch is now refused in that mode:

- Serial run, ambient `CLAUDE_CONFIG_DIR` set → honoured, unchanged. No concurrency, no hazard.
- `BENCH_PARALLEL` set → ambient value ignored, per-run temp dir injected, warning logged naming the reason.
- Unset → per-run temp dir, as before.

The parent environment is never modified in either path; the per-run value goes through `extra_env` only.

Verified in both modes: serial yields `None` (operator dir flows through `os.environ` untouched), parallel yields two distinct temp dirs across two runs, neither equal to the operator's. New test `test_execute_overrides_operator_config_dir_under_parallel` covers it.

`ruff check` / `ruff format` clean, 1254 tests passing.

eugeneng04:
@coderabbitai review

devops_bench/agents/cli/claude_code/parsing.py
```diff
@@ -0,0 +1,371 @@
...
+            # earlier good one.
+            tail = event.get("result")
+            if isinstance(tail, str) and not result_output:
+                result_output = tail
+            usage = event.get("usage")
+            if isinstance(usage, dict) and _has_usage(usage):
+                tokens = _usage_tokens(usage)
+                result_usage_seen = True
+            subtype = event.get("subtype")
+            if isinstance(subtype, str) and subtype.startswith("error_"):
```
itssimrank, line 283: Why do we have a underscore here? what if the subtype was _error?

devops_bench/agents/cli/claude_code/parsing.py
```diff
@@ -0,0 +1,371 @@
...
+        output = max(0, output - reasoning)
+    tokens.update(
+        input=_int_or_none(usage.get("input_tokens")),
+        cached=_int_or_none(usage.get("cache_read_input_tokens")),
+        cache_write=_int_or_none(usage.get("cache_creation_input_tokens")),
+        reasoning=reasoning,
+        output=output,
+    )
+    reported = [v for k, v in tokens.items() if k != "total" and v is not None]
+    if reported:
```
itssimrank: Shouldnt this also check if tokens["output"] is not None?

devops_bench/agents/cli/claude_code/parsing.py
eugeneng04: There are only 5 subtypes as listed here: https://code.claude.com/docs/en/agent-sdk/python#resultmessage, anything that gets past the prefix would still hit the is_error branch below it.

devops_bench/agents/cli/claude_code/parsing.py
eugeneng04: Yes, added that check

devops_bench/agents/cli/claude_code/agent.py
```diff
@@ -0,0 +1,390 @@
...
+# ``--mcp-config`` under ``-p`` only waits for still-pending servers to connect
+# before the first turn from this version on. An older binary accepts the flag
+# and starts the turn anyway, so an MCP-augmented arm can run with none of its
+# tools attached and still be scored as augmented — the same silent
+# contamination ``--strict-mcp-config`` prevents from the other direction.
+_MCP_WAIT_MIN_VERSION = (2, 1, 221)
+_VERSION_RE = re.compile(r"(\d+)\.(\d+)\.(\d+)")
+
+
+def _claude_version(target: str) -> tuple[int, int, int] | None:
```
itssimrank, line 111: nit: Would it make sense to cache it? Specially for when the benchmark grows to matrix of 50–100 tasks.

devops_bench/agents/cli/claude_code/agent.py
```diff
@@ -0,0 +1,390 @@
...
+            argv = _build_argv(
+                target,
+                prompt,
+                model=self.config.model,
+                max_turns=self.config.max_turns,
+                mcp_config_path=mcp_config_path,
+            )
+            with _claude_config_dir() as config_dir:
+                env_overlay = _build_env(self.config, config_dir=config_dir)
+                try:
```
itssimrank, line 344: Could you update this to also report timed out errors like other harnesses?

devops_bench/agents/cli/claude_code/parsing.py
```diff
@@ -0,0 +1,375 @@
...
+        if not isinstance(event, dict):
+            continue
+
+        etype = event.get("type")
+        if etype == "system":
+            # A failed MCP server leaves the run tool-less but still exits 0, so
+            # surface terminal-failure statuses (see ``_MCP_FAILED_STATUSES``)
+            # rather than scoring a silently-degraded run clean.
+            if event.get("subtype") == "init":
+                for server in event.get("mcp_servers") or []:
```
itssimrank: This should be done before starting iteration.

tests/unit/evalharness/test_registry_resolution.py
```diff
@@ -150,3 +166,24 @@ def test_entry_point_agent_resolves_with_no_harness_edit(
     mock_eps.assert_called_once_with(group="devops_bench.agents")
     # The harness threaded its built config into the entry-point-loaded agent.
     assert isinstance(_DummyAgent.last_config, AgentConfig)
+
+
+def test_manifest_records_the_canonical_harness_key(tmp_path, mocker: MockerFixture) -> None:
```
itssimrank: nit: Add annotations

itssimrank (approved):

devops_bench/agents/cli/claude_code/agent.py
eugeneng04: Yeah that makes sense, I added a cache of the claude version

devops_bench/agents/cli/claude_code/agent.py
eugeneng04: Yes, added a report of timed out errors

devops_bench/agents/cli/claude_code/parsing.py
eugeneng04: Changed such that mcp_servers lookup is before the loop

tests/unit/evalharness/test_registry_resolution.py
eugeneng04: Added annotations

janetkuo (approved):
/lgtm

PR #90, opened 2026-08-18
feat(agents): support multiple MCP servers, fail unreachable ones, copy whole skill bundles

eugeneng04:
Adds support for multiple MCP servers, fails a run when a granted server is unreachable, and copies whole skill bundles instead of just `SKILL.md`.

**Multiple MCP servers.** `AGENT_MCP_SERVER` holds one argv string with no env and no cwd, so only a server that reads everything from the ambient environment works. New `AGENT_MCP_CONFIG` takes the standard `mcpServers` document (the one Claude Code and Cursor read), inline or as a file path:

```json
{"mcpServers": {"github": {"command": "npx",
                           "args": ["-y", "@modelcontextprotocol/server-github"],
                           "env": {"GITHUB_TOKEN": "${GITHUB_TOKEN}"}}}}
```

`AGENT_MCP_SERVER` still works as the single-server shorthand. Secrets go in as `${VAR}` and the CLI expands them from the subprocess env — the workspace gets copied into `generated_files/`, so a rendered value would end up committed.

**Unreachable servers fail the run.** Right now a server that doesn't start is invisible: gemini falls back to `run_shell_command`, `errors` is empty, exit 0, and the record still counts as an MCP arm. Both CLI agents now check their granted servers before the model runs. gemini uses `gemini mcp list` (which also covers folder trust); openclaw has no equivalent, so it uses a small stdio probe.

**Skill bundles.** `materialize_skills` copied `SKILL.md` and dropped the `references/`, `templates/`, `scripts/` next to it, so a skill telling the agent to read `templates/report.md` pointed at nothing. It now copies the directory.

Tested against real `gemini` 0.56 and `openclaw` 2026.7.1 with two MCP servers: both models called through both servers, the untrusted-workspace and dead-server cases both failed with the reason named, and no secret reached the workspace config.

HTTP/SSE MCP servers are out of scope — the probe only speaks stdio and I had nothing remote to test against.

## Summary by CodeRabbit

- **New Features**
  - Added support for multiple MCP servers with per-server tools, environment variables, working directories, and configuration files.
  - Added MCP preflight checks across supported CLI agents, including connection validation, tool discovery, secret redaction, and actionable errors.
  - Added support for forwarding custom agent command-line flags.
  - Added API support for routing tool calls across multiple MCP servers.

- **Bug Fixes**
  - Improved MCP configuration validation and error reporting.
  - Improved skill bundle copying while safely handling symbolic links.
  - Prevented agent runs when required MCP servers are unavailable.

### DIFF
--- a/devops_bench/agents/api/agent.py
+++ b/devops_bench/agents/api/agent.py
@@ -51,11 +51,12 @@ _log = get_logger("agents.api.agent")
 
 
 class ToolNameConflictError(RuntimeError):
-    """Two granted MCP servers advertise the same tool name.
+    """One tool name is claimed twice — by two granted MCP servers, or by a
+    discovered skill and a granted server.
 
     Raised once the servers are up, since the advertised names are only known
-    then. Fatal for the run: routing the call to whichever server happened to
-    be listed first would score the arm against a toolset nobody chose.
+    then. Fatal for the run: serving whichever source happened to be resolved
+    first would score the arm against a toolset nobody chose.
     """
 
 
@@ -393,7 +394,8 @@ async def _run_async(
         per-tool dispatch failures recorded by the dispatcher.
 
     Raises:
-        ToolNameConflictError: If two granted servers advertise the same tool.
+        ToolNameConflictError: If two granted servers advertise the same tool,
+            or a discovered skill tool carries a granted server's tool name.
     """
     errors: list[str] = []
     skill_tools, skill_resources, skill_names = await asyncio.to_thread(
@@ -408,6 +410,17 @@ async def _run_async(
             for binding in mcp_bindings
         ]
         mcp_tools, routes = await _route_tools(sessions)
+        # Skill names are checked against the routed MCP names for the same
+        # reason two servers are checked against each other: the model would be
+        # handed one name twice, and the dispatcher's skill-first rule would
+        # serve the skill while the MCP tool it shadows never runs.
+        clash = sorted(routes.keys() & skill_resources.keys())
+        if clash:
+            raise ToolNameConflictError(
+                f"skill tools {clash} collide with tools advertised by a granted MCP "
+                "server; the grant is rejected rather than serving the skill and "
+                "leaving the server's tool unreachable"
+            )
         loop_result = await run_tool_loop(
             client=client,
             goal=prompt,
--- a/devops_bench/agents/api/agent.py
+++ b/devops_bench/agents/api/agent.py
@@ -22,27 +22,42 @@ entries.
 from __future__ import annotations
 
 import asyncio
+import contextlib
 import shlex
 import time
 from collections import deque
+from collections.abc import Mapping
 from pathlib import Path
-from typing import Any
+from typing import TYPE_CHECKING, Any
 
 from devops_bench.agents.api.mcp import MCPClient, extract_tool_text
 from devops_bench.agents.api.skills import (
-    SkillToolInfo,
     discover_skill_tools,
 )
 from devops_bench.agents.base import AGENTS, AgentHarness
 from devops_bench.agents.config import AgentConfig
 from devops_bench.agents.result import AgentResult, ToolCall
+from devops_bench.agents.shared.mcp_probe import child_env
 from devops_bench.core import get_logger
 from devops_bench.models import LLMClient, get_model
 from devops_bench.models.utils.loop import LoopResult, ToolDispatcher, run_tool_loop
 
-__all__ = ["ApiAgent", "fold_trajectory", "extract_tokens"]
+if TYPE_CHECKING:  # pragma: no cover - typing-only import
+    from devops_bench.agents.capabilities import McpBinding
+
+__all__ = ["ApiAgent", "ToolNameConflictError", "fold_trajectory", "extract_tokens"]
 
 _log = get_logger("agents.api.agent")
+
+
+class ToolNameConflictError(RuntimeError):
+    """Two granted MCP servers advertise the same tool name.
+
+    Raised once the servers are up, since the advertised names are only known
+    then. Fatal for the run: routing the call to whichever server happened to
+    be listed first would score the arm against a toolset nobody chose.
+    """
+
 
 # Safety cap on agent turns; overridable via ``AgentConfig.max_turns``. Set high
 # because API agents legitimately take many tool-use turns — it only guards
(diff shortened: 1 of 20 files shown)


eugeneng04:
@coderabbitai review

eugeneng04:
Fixed the UTF-8 decode finding in 8fa022b: `Path.read_text` raises `UnicodeDecodeError` (a `ValueError`), which slipped past the `except OSError` and broke `from_env`'s documented `ConfigError` contract. Added `except UnicodeError` plus a regression test. Full unit suite: 1237 passed.

eugeneng04:
@coderabbitai review

devops_bench/agents/shared/mcp_probe.py
```diff
@@ -0,0 +1,563 @@
...
+            raise McpUnreachableError("server closed its stdout before responding")
+        try:
+            message = json.loads(line)
+        except json.JSONDecodeError:
+            continue
+        if not isinstance(message, dict) or message.get("id") != request_id:
+            continue
+        if "method" in message or ("result" not in message and "error" not in message):
+            continue
+        if "error" in message:
```
itssimrank: Error should be separate in the payload that you can read from and directly use message.get("error") is not None instead.

devops_bench/agents/api/agent.py
```diff
@@ -287,51 +309,79 @@ async def dispatch(name: str, args: Any, call_id: str | None) -> str:
...
-    tools.extend(skill_tools)
-    return tools
+    routes: dict[str, MCPClient] = {}
+    owners: dict[str, str] = {}
+    for binding, mcp_client in sessions:
+        server = binding.name or "<unnamed>"
+        listed = await mcp_client.list_tools()
+        for tool in listed.tools:
+            name = getattr(tool, "name", "")
+            if name in routes:
```
itssimrank, line 352: Should this be checked against skill_resources as well?

devops_bench/agents/cli/gemini_cli/agent.py
```diff
@@ -91,6 +102,108 @@ def _build_settings(mcp_servers: tuple[McpBinding, ...], *, skills_enabled: bool
...
+        expected: Server names that must be connected.
+        env_overlay: Env overlay the run itself uses, so ``${VAR}`` references
+            in the settings resolve the same way here.
+        timeout: Seconds before the listing is abandoned.
+
+    Returns:
+        A human-readable reason, or ``""`` when every expected server is
+        connected.
+    """
+    try:
```
itssimrank, line 157: Could you add a returncode  check after the try/except block for completed so that we return failures with a syntax/config error.

devops_bench/agents/config.py
```diff
@@ -16,53 +16,243 @@
...
+            references, enforced by :func:`_reject_literal_secrets`.
+        cwd: Working directory the server is launched in; ``""`` for inherit.
+        tools: Tools to pre-approve, or ``None`` to inherit
+            ``AGENT_ALLOWED_TOOLS``. The distinction matters: an explicit ``[]``
+            grants none.
+    """
+
+    model_config = ConfigDict(strict=True)
+
+    command: str = Field(min_length=1)
```
itssimrank: nit: We should enforce non-whitespace characters using a regex pattern like pattern=r"\S"

devops_bench/agents/shared/mcp_probe.py
```diff
@@ -0,0 +1,563 @@
...
+    """Write one newline-delimited JSON-RPC message to the server's stdin.
+
+    Raises:
+        McpUnreachableError: If the server's stdin is already closed — i.e. the
+            process died on launch.
+    """
+    try:
+        proc.stdin.write(json.dumps(message) + "\n")
+        proc.stdin.flush()
+    except (BrokenPipeError, ValueError) as exc:
```
itssimrank: nit: We should also cover ConnectionResetError or other OSError.

devops_bench/agents/cli/antigravity/agent.py
```diff
@@ -277,6 +278,18 @@ def _execute(
                     json.dumps(settings, indent=2), encoding="utf-8"
                 )
 
+            # Same gate the other CLI harnesses run: a granted server that never
+            # starts leaves agy falling back to shell tools while the run is
+            # still scored as an MCP arm.
+            try:
+                preflight_mcp(
```
itssimrank, line 285: can you add a test case that does preflight_mcp args verification?

devops_bench/agents/shared/mcp_probe.py
eugeneng04: Added

devops_bench/agents/cli/gemini_cli/agent.py
eugeneng04: Added with a cap of 800 chars so we don't have a super long output

devops_bench/agents/api/agent.py
eugeneng04: Yeah, added a check against skill_resources and raises `ToolNameConflictError`

devops_bench/agents/config.py
eugeneng04: Added enforcement of non-whitespace characters

devops_bench/agents/shared/mcp_probe.py
eugeneng04: Added OSError and ValueError which covers ConnectionResetError

devops_bench/agents/cli/antigravity/agent.py
eugeneng04: Yes added a test case

itssimrank, line 285: Also curious since every server is launched twice per run (probe, then the real launch), can this cause trouble for launchers that cannot tolerate that?

devops_bench/agents/config.py
```diff
@@ -16,53 +16,245 @@
...
+        env: The server's declared ``env`` mapping.
+
+    Raises:
+        ConfigError: If a secret-named key holds a value with no ``${VAR}``
+            reference in it.
+    """
+    for key, value in env.items():
+        if not value or _ENV_REF.search(value):
+            continue
+        if any(hint in key.upper() for hint in _SECRET_KEY_HINTS):
```
itssimrank, line 67: This needs to be a boundary-aware check, otherwise we'll have a lot of false positives.

itssimrank (approved):

PR #115, opened 2026-08-20
Add a vcluster infrastructure provider for fast task iteration

geojaz:
## What this adds

A `vcluster` infrastructure provider that provisions [loft-sh/vcluster](https://github.com/loft-sh/vcluster) virtual clusters inside an existing host cluster, instead of creating a real GKE cluster per run. Cluster provisioning drops from 10-20 minutes to 2-4 minutes, and workloads inside the vcluster schedule in seconds. That changes how fast we can iterate while developing tasks.

Tasks opt in with `provider: "vcluster"` (or `INFRA_PROVIDER=vcluster` at run time). No task schema changes.

## What's in the change

- `tf/modules/cluster/vcluster`: the vcluster submodule, alongside `gke` and `kind`. It creates a LoadBalancer Service before the Helm release, waits for the ingress address, and bakes that address into the proxy cert SANs and the exported kubeconfig. Access needs no vcluster CLI: the kubeconfig comes from the `vc-<name>` secret and is written to a per-cluster path.
- The namespace is managed by tofu rather than Helm, so destroy removes the PVC. Without this, a re-created vcluster silently resumes the previous run's state (we hit this in testing).
- `tf/prebuilt/vcluster`: standalone stack. `tf/prebuilt/minimal`: a provider-agnostic stack that dispatches to `tf/modules/cluster` for gcp, kind, or vcluster. Named `minimal` rather than `minimum` to avoid colliding with downstream stacks of that name.
- `devops_bench/providers/vcluster.py`: `VclusterProvider`, mirroring the KinD provider shape. Handles host cluster credentials (gcloud on GKE hosts) and derives per-cluster kubeconfig paths so parallel runs don't collide.
- Docs in `docs/components/infra.md`, 20+ unit tests.

## Host clouds

The module is host-cloud agnostic: a `host_cloud` variable (`gke`, `eks`, `aks`; default `gke`) is threaded through, and the LB-address approach works on all three (IP on GKE/AKS, NLB hostname on EKS). GKE is implemented and tested. For EKS/AKS the remaining work is host credential automation (`aws eks update-kubeconfig` / `az aks get-credentials`) and a real test pass on each; until then those paths assume the host context already exists in kubeconfig.

## When to use it (and not)

vcluster is right for workload, manifest, and policy-level tasks, which is most of them. It is not faithful for anything that depends on real node behavior: node pools, per-cluster GKE Workload Identity, node-level DaemonSets, or Autopilot scaling semantics. Those tasks should keep using the `gcp` provider.

## Verification

Tested live against a GKE Autopilot host cluster (chart v0.36.1):

| Phase | Time |
|---|---|
| `tofu apply` to responsive vcluster API | 2m17s - 3m36s (varies with Autopilot node warmth) |
| nginx deployment inside the vcluster to Ready | 3-4s |
| `tofu destroy` (namespace, PVC, LB fully removed, verified) | 68-86s |

Full unit suite passes (1190 tests). `tofu validate` and `fmt` clean across `tf/`.

## Summary by CodeRabbit

- **New Features**
  - Added support for provisioning and managing vcluster virtual clusters on GKE, EKS, and AKS.
  - Added configurable host-cluster contexts, kubeconfig paths, locations, projects, and chart versions.
  - Added cluster outputs for names, locations, endpoints, and kubeconfig files.
  - Added minimal and dedicated vcluster deployment configurations.

- **Documentation**
  - Documented vcluster setup, required configuration, provisioning, teardown, prerequisites, and host-cloud limitations.

- **Tests**
  - Added coverage for provider registration, configuration defaults, credential handling, and cluster output resolution.

### DIFF
--- a/devops_bench/providers/vcluster.py
+++ b/devops_bench/providers/vcluster.py
@@ -1,0 +1,130 @@
+# Copyright 2026 The Kubernetes Authors.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""vcluster provider: virtual clusters running inside an existing GKE host."""
+
+from __future__ import annotations
+
+from pathlib import Path
+from typing import Any
+
+from devops_bench.core import ClusterInfo, ConfigError, get_env, get_logger
+from devops_bench.core.subprocess import run
+from devops_bench.providers.base import PROVIDERS, Provider, ResolveContext
+
+__all__ = ["VclusterProvider"]
+
+_log = get_logger("providers.vcluster")
+
+
+@PROVIDERS.register("vcluster")
+class VclusterProvider(Provider):
+    """Provider for loft-sh vcluster virtual clusters hosted on GKE.
+
+    A vcluster is a virtual Kubernetes control plane running as a workload
+    inside an existing GKE host cluster, provisioned in a couple of minutes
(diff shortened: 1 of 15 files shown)


geojaz:
Sorry- i didn't see #77 or #76 🙄 ... my bad. i think i'll probably close this and provide my feedback on those PRs

PR #124, opened 2026-08-21
Add the operational skills and references

pradeepvrd:
Agent skills for the day-to-day eval workflow — `run-eval` (single runs), `run-parallel-evals` (matrix runs), `diagnose-eval-failure` (working back from a bad score), and `docs-sync` (keeping docs honest as code changes) — plus references covering the matrix runner, monitoring and recovery, and unlimited mode.

For permission setup, the tool-neutral guidance in `.agents/skills/devops-bench-review/SKILL.md` remains the source of truth; no sample configs are shipped.

## Summary by CodeRabbit

* **New Features**
  * Added guidance for running individual and parallel evaluations locally or remotely.
  * Added monitoring, recovery, failure diagnosis, and opt-in self-healing procedures.
  * Added documentation synchronization guidance for keeping related documentation current.
  * Added safeguards for shared-environment changes, retries, isolation, and accurate result reporting.

* **Documentation**
  * Documented authentication, dry runs, detached execution, result locations, matrix configuration, and operational safeguards.
  * Documented globally unique task-directory naming requirements to prevent run and infrastructure collisions.
  * Clarified infrastructure-flake handling, fresh retry launches, and routing for model, infrastructure, and rubric issues.

PR #128, opened 2026-08-24
Enforce unique task directory basenames in review guidelines

pradeepvrd:
The matrix runner (`scripts/bastion/run_matrix.sh`) derives run IDs, cluster names, and output directories from the task directory basename, not `task_id`, so two task directories with the same folder name (e.g. `tasks/gcp/foo` and `tasks/kind/foo`) would share infrastructure identity even with distinct ids.

Rather than collision-proofing the runner, this encodes basename uniqueness as an authoring invariant, enforced where reviews happen:

- `.coderabbit.yaml`: add the rule to the `tasks/**` path instructions so it fires automatically on every PR that adds or renames a task directory.
- `.agents/skills/task-review`: mirror the same check in the schema/metadata checklist, next to the existing `task_id` uniqueness check.

No collisions exist in the tree today.

Addresses the unresolved CodeRabbit finding on #124 (`.agents/references/running-evals.md`).

## Summary by CodeRabbit

* **Documentation**
  * Clarified that task directory names must be globally unique across all task paths.
  * Documented that directory names determine run IDs, cluster names, and output directories.
  * Added review guidance to identify naming collisions during task validation.

### DIFF
--- a/.agents/skills/task-review/SKILL.md
+++ b/.agents/skills/task-review/SKILL.md
@@ -45,9 +45,8 @@ silently dropped). `task_id` is **globally unique** (`grep -rn '^task_id\|^id:'
 tasks/`). The task's **directory basename is globally unique** across `tasks/**`
 (compare parent dir names of every `tasks/**/task.yaml`) — the matrix runner keys
 run IDs, cluster names, and output dirs on the folder name, not `task_id`, so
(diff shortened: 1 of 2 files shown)


pradeepvrd:
Folded into #124, where the finding was raised.

Issue #97, opened 2026-08-19
feat(monitor): Add Performance Monitor to capture resource metrics across tasks


AishSundar:
### Problem / Motivation
Beyond binary pass/fail outcomes, understanding the efficiency of agent execution (token consumption, latency, API call count, resource overhead) is critical for comparing models and practical deployment feasibility.

### Proposed Solution & Technical Details
- Implement a Performance Monitor module that wraps task execution.
- Capture telemetry:
  - Total input/output token usage and cost estimations.
  - Step count and wall-clock execution latency per task.
  - Number of tool / kubectl invocations and command execution errors.
  - Host CPU / memory utilization during benchmark runs.
- Output metrics into structured evaluation summary artifacts.

### Acceptance Criteria
- [ ] Performance monitor captures token, step, and latency metrics across all executed tasks.
- [ ] Structured performance statistics included in the final evaluation report JSON.
- [ ] Negligible overhead introduced to task execution runtime.

---
/kind feature
/area infra
/priority critical-urgent

eugeneng04:
Opened #129, which covers the per-task telemetry half of this.

**What is added**, all on `ResultRow`

| Item | Status |
|---|---|
| Token usage | Six buckets: `input`, `cached`, `cacheWrite`, `reasoning`, `output`, `total` — where the total is the sum of the rest. Also fixes two bugs that were zeroing buckets: the normalizer only read snake_case keys while the CLI harnesses emit camelCase, and openclaw's session rollup drops `cacheWrite` entirely (recovered from the per-call events). |
| Step count | `modelTurns` (model round-trips) and `toolCalls`, kept separate — one turn can issue several tool calls and a text-only turn issues none. |
| Wall-clock latency | `latencySec` is now the **agent span** only; workspace setup and transcript parsing used to be inside it. Paired with `timeoutSec` (the budget the iteration ran under), since "timed out" and "used 90% of its budget" are uninterpretable without it, and `toolWaitSec` (how much of the latency was waiting on tools, concurrent calls counted once) so a cold cluster and a slow model aren't the same number. |
| Tool / command errors | `toolErrors`. The previous count included successful calls. |
| Structured output | All of the above plus `terminalReason` (`completed` / `timeout` / `error` — distinct from `status`, which describes the record) and `servedModel` (the model the provider actually answered with, which isn't always the one requested). Documented in `docs/components/metrics.md`. |
[...]

PR #129, opened 2026-08-24
feat(results): record per-run telemetry on every result row

eugeneng04:
Addresses the telemetry half of #97 — the per-run resource metrics the leaderboard cannot recompute after the fact.

## What lands on `ResultRow`

| Field | Type | Why the dashboard cannot derive it |
|---|---|---|
| `terminalReason` | `str` | Why the *agent* stopped (`completed` / `timeout` / `error`). Distinct from `status`, which describes the record: a run killed at its budget still reads `status: "success"`. |
| `toolCalls` | `int \| None` | The unit of agentic work. Two models with the same score and the same wall clock can differ several-fold here, and the trajectory itself is too large to aggregate at dashboard time. |
| `toolErrors` | `int \| None` | A high count against a passing score means the model recovered; against a failing one it usually means the environment broke, not the model. |
| `modelTurns` | `int \| None` | Model round-trips. Not `toolCalls` — one turn can issue several tool calls, and a text-only turn issues none. |
| `toolWaitSec` | `float \| None` | How much of `latencySec` was spent waiting on tools, concurrent calls counted once. Without it, a cold cluster and a slow model are the same number on a leaderboard that ranks latency lower-is-better. |
| `timeoutSec` | `float \| None` | The wall-clock budget the iteration ran under. "Timed out" and "used 90% of its budget" are both uninterpretable without it. |
| `servedModel` | `str` | The model the provider actually answered with, which is not always the one requested (aliases, fallbacks, silent version bumps). |

Every field rides on the **row**, not the manifest: ingest uploads `rows.json` alone and never reads `manifest.json`, so a run-level setting only reaches the dashboard by riding along on each row.

## Harness-side fixes these depend on

- **Latency is the agent span, not the harness run.** `latencySec` previously included workspace setup and trajectory parsing. It now brackets the subprocess only.
- **OpenClaw cache-write tokens recovered.** The `model.completed` session rollup omits `cacheWrite` from both its buckets and its sum; the per-call `assistant.message.usage` is the only event carrying it. That one bucket is summed from there, folded back into the total, and every other bucket comes from `model.completed` — so nothing is double-counted and the buckets still reconcile against the total.
- **CLI camelCase token keys read.** The CLI harnesses emit `cacheRead` / `cacheWrite`; the normalizer only looked for snake_case, so those buckets normalized to `None` on every CLI run.
- **Concurrent tool calls merged, not summed.** OpenClaw can issue two `tool.call` events at the same millisecond from one `assistant.message`. `agents/shared/timing.py` takes the interval union so `toolWaitSec` can never exceed `latencySec`.
- **Tool errors counted, not tool calls.** The previous count included successful calls.
- **The ADK harness reports the same telemetry.** It landed on `main` after this branch was cut and was the one harness still writing rows with nothing on them. `model_version` -> `servedModel`, the events carrying `usage_metadata` -> `modelTurns` (exactly the events the token sums are built from, so the two agree), and the `function_call` / `function_response` event timestamps -> `toolWaitSec`. Its `_drive` now returns *why* it stopped rather than leaving the caller to match on an error string, and its latency brackets the run rather than the whole `_execute`.
- **Partial telemetry recovered from timed-out runs.** A killed openclaw run leaves a complete session in `OPENCLAW_STATE_DIR`, and a killed gemini run leaves a valid prefix of its stream-json. Both are now parsed, so a timed-out row still reports how far the agent got instead of reporting nothing.

## Conventions held throughout

- **Unreported is `None`, never `0`.** Applied per field, not blanket: `modelTurns` drops a `0` (a run that produced output cannot have taken zero round-trips), `toolWaitSec` keeps a genuine `0.0` (tools can return inside the transcript's millisecond resolution), `servedModel` collapses to `""`. All three reject `bool`, since `True` is an `int` in Python.
- **Six canonical token buckets** — `input`, `cached`, `cacheWrite`, `reasoning`, `output`, `total`, where the total is the sum of the rest.
- No change to `SCHEMA_VERSION`: every field is additive with a default, so existing rows stay valid.

## Validation

`1459 passed` (`1448 passed, 11 skipped` without the optional `adk` extra, which is what CI installs), `ruff check` clean. Each field is covered at three layers: parser (per harness), `AgentResult`, and `build_rows`.

End-to-end against a real run (`devops-bench --agent-type adk --no-infra`, stub-model agent, tool sleeping 0.4s), the row that reached `rows.json`:

```
servedModel    'stub-model-001'   # not 'adk', the id that was requested
terminalReason 'completed'
timeoutSec     60.0
toolCalls      1     toolErrors 0     modelTurns 2
latencySec     0.657               toolWaitSec 0.409
```

and the same run under a 2s budget it blows:

```
terminalReason 'timeout'   status 'success'   timeoutSec 2.0
toolCalls      1           modelTurns 1       totalTokens 110
```

— the distinction `status` alone cannot make, with the partial telemetry intact.

`docs/components/metrics.md` documents the new fields under the `rows.json` contract.

## Follow-up, not in this PR

Each new column still needs a matching `load.mjs` validator, a `derive()` projection, and metric registration on the dashboard side before it is visible.

## Summary by CodeRabbit

* **New Features**
  * Added richer run telemetry, including terminal status, latency, model turns, served models, tool usage, tool wait time, and timeout budgets.
  * Preserved partial output and telemetry when command-line runs time out or encounter errors.
  * Improved token accounting across supported providers, including cache usage.
* **Documentation**
  * Documented new result fields and clarified latency and token reporting.
* **Bug Fixes**
  * Timeout failures are now distinguished from other subprocess errors.
  * Concurrent tool durations are merged accurately to avoid double counting.


eugeneng04:
Thanks — worked through all of these. Five fixed, two answered.

**Fixed**

- **Preserve known zero tool counts.** Done without a contract change: the record already carries the tiebreaker. Every path that loses a transcript reports why in `errors` — the openclaw exporter appends its failure, a timeout appends its own, and a failed record carries the exception — so an empty trajectory beside an *empty* `errors` list is a genuine `(0, 0)`, and anything else stays `None`. This is the same test the harness applies to `validated` (`not errors and bool(trajectory)`), minus the part that discards a clean zero.
- **Count only canonical tool-call entries.** Filtered on a string `name`. Added a regression test with a text turn interleaved between two calls.
- **Measure latency before credential cleanup.** `agent_sec` is now stamped at the top of the `finally`, before `copied_token.unlink()`.
- **Limit the total-token guarantee to canonical telemetry.** Correct — `normalize_tokens` passes pre-canonical totals through unchanged and those can exclude cached/reasoning. Reworded.
- **Exercise the timeout path in that test.** Good catch; the `SubprocessError` was missing `timed_out=True`, so it would have passed even if timeouts were labelled `error`. Set it and added the `terminal_reason` assertion.

**Not changing**

- **Vendor neutrality in `result.py` / `timing.py`.** `AGENTS.md` scopes this rule to the *cloud* axis and says so explicitly: "a *cloud provider* (`gcp`, `kind`) provisions the cluster, a *model provider* (`gemini`, `claude`, `ollama`) serves the LLM — this rule is about the cloud axis." `openclaw` and `gemini` are model-axis names, and both docstrings cite a concrete observed artifact (an alias resolution seen in a real run; two `tool.call` events stamped at the same millisecond from one message) — the exemption the same rule grants for "where they name a real provider artifact". Generalising them would remove the evidence for why the code is shaped that way. Happy to reword if maintainers read the rule differently.
- **Enforcing `TERMINAL_REASONS` in `AgentResult`.** Validating at construction turns a telemetry typo into a failed benchmark run, and the field is diagnostic rather than functional. Coercing in `normalize.py` instead would create a `results` → `agents` import, which this codebase deliberately avoids — `core/score_keys.py` exists for exactly that reason ("keeps one definition without creating a `metrics` <-> `results` edge"). Every producer's value is pinned by a test today. If maintainers want it enforced, the clean way is to move the constant into `core/` alongside `score_keys`; say the word and I'll do that here.
- **Type annotations on the new tests.** The three files involved (`test_agents_cli_antigravity.py`, `test_subprocess.py`, `test_results_normalize.py`) are unannotated throughout — none of their existing top-level tests carry `-> None`, and zero nested helpers in the CLI test files are annotated. Annotating only the new additions would make those files less internally consistent, not more. The files that *do* annotate (`test_agents_result.py`, `test_agents_shared_timing.py`, `test_default_harness.py`) got annotated additions.

`ruff check` clean, 1234 tests passing.

results-e2e/run_20260915_232015_092438/generated_files/k8s/gemma3-12b/00-namespace.yaml
isadominguez314: Remove this local run output from the PR I don't think it makes sense to have it. I would remove the entire results-e2e/run_20260915_232015_092438/ folder from the commit.

tasks/noop/create-deployment/task.yaml
```diff
@@ -0,0 +1,160 @@
+task_id: 2
+name: "create-deployment"
```
isadominguez314: If you want to add this new task, I would separate it into a different PR especially because this PR asserts that there is no behavior change. I'd remove this from PR as well. Also for future reference, the task uses GKE/GCP language and is under tasks/noop/ so it's not vender neutral.

devops_bench/agents/cli/claude_code/parsing.py
```diff
@@ -214,15 +225,17 @@ def parse_stream_json(stdout: str) -> tuple[str, list[dict], dict, list[str]]:
...
+            # A CLI-rendered error envelope carries a synthetic model and no real call.
+            if not event.get("is_api_error_message"):
+                # Per-turn usage backs a truncated stream; dedupe the repeated id.
+                msg_id = message.get("id")
+                identified = isinstance(msg_id, str) and bool(msg_id)
+                if not (identified and msg_id in seen_message_ids):
+                    if identified:
+                        seen_message_ids.add(str(msg_id))
+                    _add_usage(acc_usage, message.get("usage"))
+                note_model(served_models, message.get("model"))
```
isadominguez314: Here `served_models` and `model_turns` count every assistant message including subagent messages (ones tagged with `parent_tool_use_id`). If the agent starts a task-tool subagent on a different model, the row reports `servedModel = "<main>,<helper>"`. The row docs treat that as a sign the run switched models, but it didn't. The subagent's turns also inflates `modelTurns`. To fix, skip messages that have `parent_tool_use_id` set.

devops_bench/agents/adk/agent.py
```diff
@@ -476,28 +477,22 @@ async def _close_quietly(runner: Any) -> None:
         await closing
 
 
-def _drive(root_agent: Any, prompt: str, timeout_sec: float | None) -> tuple[list[dict], list[str]]:
+def _drive(
+    root_agent: Any, prompt: str, timeout_sec: float | None
+) -> tuple[list[dict], list[str], str]:
     """Run the agent to completion and collect its serialized event stream.
```
isadominguez314: I know you intentionally shortened all of these comments but I think at least having args/returns documentation is valuable. This also matches the other private helpers in the file. Same applies to the `_fold_response` in adk/parsing.py.

devops_bench/agents/cli/antigravity/agent.py
```diff
@@ -49,32 +49,21 @@
 _UNDECODABLE_MAX_ATTEMPTS = 2
 
 
-def _read_db_tokens(db_path: pathlib.Path) -> dict | None:
-    """Read canonical token usage from the conversation DB.
-
-    Polls for the async flush while the DB reports ``pending``.
-
-    Args:
-        db_path: Path to the ``conversations/<uuid>.db`` file.
```
isadominguez314: Same thing with these comments and throughout this PR just choose how detailed you want to be either by adding the args/returns docs back or not.

PR #137, opened 2026-08-25
feat(agents): add an ADK agent harness

jessie1111101:
## What

Adds a fifth agent harness, registered as `--agent-type adk`, that runs any agent built with the [Agent Development Kit](https://google.github.io/adk-docs/) against a benchmark task and recovers a scoreable trajectory from it.

```bash
uv sync --extra adk

export BENCH_AGENT_TYPE=adk
export AGENT_TARGET=~/agents/my_agent   # or my_pkg.agent:root_agent
export AGENT_MODEL=gemini-2.5-pro       # optional; unset keeps the agent's own model
```

The harness imports the agent named by `AGENT_TARGET`, deep-copies it, applies the benchmark's capabilities (MCP toolset, skills, rules), drives it in-process through ADK's `Runner`, and folds the `function_call` / `function_response` parts of the event stream into a trajectory.

## Why in-process rather than A2A

A2A can issue a prompt to a deployed agent but exposes no trajectory, and trajectory is most of what the benchmark scores. ADK's `Runner` event stream exposes both the answer and the tool calls, so the harness drives the agent in-process instead of POSTing to a deployed service. That is a deliberate scope boundary — see *Known limitations* below.

## Design notes

**Deep copy per run.** The imported agent is `model_copy(deep=True)`-ed before every run, so the harness's edits — model override, appended instruction text, attached toolsets — never mutate the module that a second run re-imports.

**Capabilities apply to the whole tree.** The model override, the MCP toolset, the rules brief, and discovered skills are applied to *every* agent in the tree, not just the root. ADK resolves tools from whichever agent is **active** (`LlmAgent.canonical_tools`, no inheritance from a parent), so a root-only grant leaves a delegate unable to touch the cluster. This was a real defect, found by pointing the harness at a delegating tree — see *Testing*.

Two consequences worth a reviewer's eye:

- A coordinator its author deliberately left toolless **will** receive the MCP toolset. That's the intended trade: an agent holding a tool it doesn't need is recoverable, an acting agent with no cluster access is not.
- One MCP binding is still one server process. The same toolset object is shared across the tree rather than rebuilt per agent (`McpToolset.close` is documented as safe to call more than once, which is what ADK's cleanup does when a toolset is reachable from several agents).

Workflow agents (`SequentialAgent` and friends) hold no instruction or tools of their own and are stepped over. An agent using a *callable* instruction provider can't be appended to; only that agent is skipped, and it is named in the result's errors rather than dropped silently. `AgentResult.metadata` records what landed — `model_override_count`, `instruction_agents`, `mcp_agents`, `mcp_toolsets`. Note these live on the in-process result only: the orchestrator's `results.json` carries no harness metadata today. That's pre-existing and harness-agnostic, not introduced here; the last commit corrects the doc wording rather than widening this PR to change the result record.

**Import hygiene.** All `google.adk` imports are function-local. `import devops_bench.agents.adk.agent` pulls neither `google.adk` nor `mcp`, enforced by a subprocess test — the SDK is an optional extra and must stay one.

**The trajectory parser is SDK-free.** `parsing.py` operates on serialized event mappings, so the bulk of the logic is testable without the extra installed. 9 of the 43 new tests need the SDK and skip cleanly without it.

**Workspace as cwd.** The run happens with the harness-owned workspace as the process working directory, matching the `cwd` the CLI harnesses give their subprocess. See the second commit — this was a real defect found in e2e, not a precaution.

## Provider contract

`adk` sits outside the shared `AGENT_PROVIDER` contract on purpose: model routing belongs to ADK, which resolves a model string and its credentials itself. `AGENT_PROVIDER` and `AGENT_API_KEY` are **not consumed**. Authenticate the way ADK expects. Running on Vertex rather than AI Studio needs:

```bash
unset GOOGLE_API_KEY GEMINI_API_KEY     # either one silently wins over the Vertex switch
export GOOGLE_GENAI_USE_VERTEXAI=true
export GOOGLE_CLOUD_PROJECT=my-project
export GOOGLE_CLOUD_LOCATION=global
```

This is documented in `docs/components/agents.md`. Mapping the bench's `AGENT_*` contract onto ADK's `GOOGLE_*` automatically is a reasonable follow-up; I left it manual and documented rather than guessing at the mapping.

## Testing

**Unit:** 43 new tests. Full suite passes both ways — `1255 passed` with `--extra adk`, `1246 passed, 9 skipped` without it. `ruff check`, `ruff format --check`, `hack/boilerplate.py`, and `pre-commit run --all-files` all clean.

**Single-agent, end to end** on a real cluster:

1. Smoke test — probe agent, no cluster: import → deep copy → model override → `Runner` → event stream → trajectory.
2. `tasks/common/opa-remediation` on kind, with a live `gke-mcp` toolset: 30 tool calls, `status: success`, `errors: []`, `VerificationCoverage: 1.0`, `OutcomeScore: 0.8165`.
3. Same task re-run after the workspace fix, confirming the fix and no regression — identical scores, `generated_files/report.md` now collected.
4. A forced MCP probe against the live server, to check the result-envelope unwrapping against a real response rather than a recorded fixture.

**Multi-agent tree.** The runs above used a single `LlmAgent`, which cannot distinguish a root-only grant from a tree-wide one. Re-running against a three-node delegating tree (toolless coordinator → cluster operator → reporter, with callbacks recording ground truth independently of the harness) exposed the defect the last commit fixes:

| | before | after |
| --- | --- | --- |
| operator sub-agent's tools | `[]` | `["McpToolset"]` |
| operator sub-agent got the rules brief | no | yes |
| real MCP calls made by the sub-agent | 0 | 6 |
| reporter's own tools preserved | — | `["function", "McpToolset"]` |

The pre-fix run reported `status: success` with `errors: []` and scored `0.0` — a harness fault was indistinguishable from an agent that simply failed the task, which is why the fix ships with tests that assert the tree shape rather than just the root.

Also confirmed by these runs, previously untested: delegation itself, the workspace `cwd` holding across a two-hop delegation (a sub-agent wrote a relative path and the orchestrator collected it), and callbacks firing without breaking the harness.

**Scoring parity.** A delegating tree then ran the same `opa-remediation` task the single agent did, start to finish:

| | single `LlmAgent` | three-node tree |
| --- | --- | --- |
| `OutcomeScore` | 0.8165 | **0.8165** |
| `VerificationCoverage` | 1.0 | 1.0 |
| `status` / `errors` | success / `[]` | success / `[]` |
| trajectory entries | 30 | 30 (26 `run_shell` + 3 `transfer_to_agent` + 1 `write_file`) |

Same score, full verification coverage, so a tree is graded exactly like a single agent. (The matching entry counts are coincidence — the tree spends 3 of its 30 on delegation hops.) The tree split the work as intended — the operator sub-agent made all 26 cluster calls and handed off, the reporter wrote the deliverable — and both sub-agents held their own tools *plus* the attached toolset (`["function", "McpToolset"]`), confirming the grant lands alongside rather than replacing. The one failed judge check was the same one the single agent missed (an omitted policy-enforcement detail in the report), which is agent quality, not harness behaviour.

That run also used a quarantined `HOME`, because earlier runs of this task family had left `opa-repo-*.git` and a `remediation-report.md` in the real home directory — where this task provisions its own repo. I audited all 26 shell commands afterwards: every one is legitimate task work, no attempt to read the task definition or rubric.

I used throwaway probe agents rather than a real one, with tools of their own, so that any failure was unambiguously the harness's and so harness-attached toolsets had to land *alongside* existing tools.

The one judge check that failed in (2) and (3) was the agent omitting a policy detail from its report — agent quality, not harness behaviour.

## Lockfile churn

Declaring the extra moves 246 lines of `uv.lock`, because uv resolves extras universally — declaring `adk` folds `google-adk`'s constraints into the single shared resolution even for people who never install it. Net effect: `google-auth` 2.55.1→2.57.0 and `google-genai` 2.10.0→2.19.0 up; `opentelemetry-api`/`-sdk` 1.43.0→1.42.1, `opentelemetry-semantic-conventions` 0.64b0→0.63b1, and `websockets` 16.0→15.0.1 down. Full suite passes on the new resolution. This is the same trade-off the existing `anthropic` and `openai` extras already make, but flagging it explicitly since the downgrades are the kind of thing worth a reviewer's eye.

## Known limitations

Deliberately out of scope, listed so nobody assumes otherwise:

- **In-process only.** An agent that exists solely as a deployed A2A service cannot be evaluated by this harness.
- **Sub-agent attribution is flattened.** Capabilities now reach every agent, but the *trajectory* still records tool calls without which agent made each one, so `transfer_to_agent` hops and a delegate's own calls sit side by side. Invisible for a single `LlmAgent`; cosmetic but real for a tree.
- **`AGENT_MAX_TURNS` is not wired.** It remains api-harness-only. ADK's own `RunConfig.max_llm_calls` default of 500 is the effective backstop.
- **Non-Gemini models are not reachable out of the box.** ADK's `LiteLlm` wrapper needs `google-adk[extensions]`, which the `adk` extra does not install (`adk = ["google-adk>=2.7.1"]`; no `litellm` in the lock). I verified `LiteLlm` imports and constructs fine *with* that extra, so this is a packaging choice we can revisit, not a design limit — left out here to keep the lockfile churn to what the core needs. Token accounting also assumes genai usage field names, so a `LiteLlm`-backed run may report usage incompletely.
- **Harness metadata is not persisted.** `results.json` carries no `metadata`, so the `mcp_agents` / `instruction_agents` counts are in-process only. Pre-existing and not ADK-specific.

I'd rather land the core and let real usage decide which of these matter than pre-emptively guess.

## Related

- #136 caps `mcp<2`. Independent, but relevant: ADK's `McpToolset` caps itself at `mcp<2` only under extras we don't request, so this PR would otherwise inherit no ceiling.

- Tracked by #138 (ADK support, end to end).

## Summary by CodeRabbit

* **New Features**
  * Added an optional ADK harness for running Agent Development Kit agents in-process.
  * Supports configurable models, instructions, skills, MCP tools, multi-agent trees, timeouts, and workspace isolation.
  * Captures assistant responses, tool-call trajectories, token usage, and execution errors.
  * Supports flexible agent target formats and preserves partial results when runs encounter failures.

* **Documentation**
  * Added setup, configuration, target resolution, and behavior guidance for the ADK harness.

* **Tests**
  * Added comprehensive coverage for parsing, configuration, execution, tool handling, errors, and optional dependency behavior.

### DIFF
--- a/devops_bench/agents/adk/agent.py
+++ b/devops_bench/agents/adk/agent.py
@@ -234,6 +234,52 @@ def _append_instruction(agent: Any, text: str) -> bool:
     return True
 
 
+def _apply_instruction(agent: Any, text: str) -> tuple[int, list[str]]:
+    """Append ``text`` to every instruction-bearing agent in the tree.
+
+    Rules and skills are granted to the *run*, not to one node of it. The
+    benchmark grades a tree as a single agent, and a delegate that never sees
+    the operator brief can violate a constraint the root was told about.
+
+    Workflow agents (``SequentialAgent`` and friends) carry no instruction and
+    are skipped. A node using a callable instruction provider is named in the
+    returned list, so the caller can report it rather than drop it silently.
+    """
+    delivered = 0
+    refused: list[str] = []
+    for node in _iter_agents(agent):
+        if not hasattr(node, "instruction"):
+            continue
+        if _append_instruction(node, text):
+            delivered += 1
+        else:
+            refused.append(str(getattr(node, "name", None) or "<unnamed>"))
+    return delivered, refused
+
+
+def _attach_toolsets(agent: Any, toolsets: list[Any]) -> int:
+    """Attach ``toolsets`` to every tool-bearing agent in the tree.
+
+    ADK resolves tools from the *active* agent alone (``BaseLlmFlow`` asks
+    ``LlmAgent.canonical_tools``, which returns that agent's own ``tools``);
+    nothing is inherited from a parent. Attaching to the root only would leave
+    every delegate unable to touch the cluster the benchmark provisioned, and
+    the run would still report success — it would just score zero.
+
+    The toolset objects are shared across nodes rather than rebuilt per node,
+    so a binding still means one MCP server process. ``McpToolset.close`` is
+    documented as safe to call more than once, which is what ADK's cleanup does
+    when the same toolset is reachable from several agents.
+    """
+    attached = 0
+    for node in _iter_agents(agent):
+        if not hasattr(node, "tools"):
+            continue
+        node.tools = list(getattr(node, "tools", None) or []) + toolsets
+        attached += 1
+    return attached
+
+
 def _render_skills(paths: tuple[str, ...]) -> tuple[str, list[str]]:
     """Render discovered skills as an instruction section.
 
@@ -468,16 +514,20 @@ class AdkAgent(base.AgentHarness):
         if skills_section:
             sections.append(skills_section)
             metadata["skills"] = skill_names
-        if sections and not _append_instruction(prepared, "\n\n".join(sections)):
-            errors.append(
-                "agent uses a callable instruction provider; rules and skills were not delivered"
-            )
+        if sections:
+            delivered, refused = _apply_instruction(prepared, "\n\n".join(sections))
+            metadata["instruction_agents"] = delivered
+            if refused:
+                errors.append(
+                    "rules and skills were not delivered to agents using a callable "
+                    f"instruction provider: {', '.join(refused)}"
+                )
 
         toolsets, toolset_errors = _build_toolsets(self.mcp_servers)
         errors.extend(toolset_errors)
         if toolsets:
-            prepared.tools = list(prepared.tools or []) + toolsets
             metadata["mcp_toolsets"] = len(toolsets)
+            metadata["mcp_agents"] = _attach_toolsets(prepared, toolsets)
 
         return prepared, metadata, errors
 
--- a/devops_bench/agents/adk/agent.py
+++ b/devops_bench/agents/adk/agent.py
@@ -1,0 +1,489 @@
+# Copyright 2026 The Kubernetes Authors.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""In-process harness driving an agent built with the Agent Development Kit."""
+
+from __future__ import annotations
+
+import asyncio
+import importlib
+import importlib.util
+import json
+import os
+import pathlib
+import sys
+from types import ModuleType
+from typing import TYPE_CHECKING, Any
+
+from devops_bench import core
+from devops_bench.agents import base
+from devops_bench.agents import config as agents_config
+from devops_bench.agents import result as agents_result
+from devops_bench.agents.adk import parsing
+from devops_bench.agents.shared import cli_capabilities
+from devops_bench.agents.shared import skills as shared_skills
+
+if TYPE_CHECKING:
+    from devops_bench.agents import capabilities
+
+__all__ = ["AdkAgent"]
+
+_log = core.get_logger("agents.adk")
+
+#: Optional dependency group carrying the ADK SDK.
+_EXTRA = "adk"
+
+#: Attribute an ADK agent module is expected to expose, matching the convention
+#: ADK's own loader follows.
+_ROOT_AGENT_ATTR = "root_agent"
+
+#: Submodule searched when a target package does not expose the root agent
+#: itself — again ADK's own ``<agent_dir>/agent.py`` convention.
+_AGENT_SUBMODULE = "agent"
+
+#: Identifiers for the throwaway session each run drives the agent through.
+_APP_NAME = "devops-bench"
+_USER_ID = "devops-bench"
+
+#: Budget for an MCP server to start and list its tools. Deliberately generous:
+#: a tool server that shells out to a cloud CLI can be slow on a cold start, and
+#: the SDK default is tuned for local in-memory servers.
+_MCP_STARTUP_TIMEOUT_SEC = 60.0
+
+
+def _looks_like_path(spec: str) -> bool:
+    """Report whether a target names a filesystem location, not a module.
+
+    A dotted module path (``my_pkg.agent``) and a filesystem path are both
+    plausible targets, so the two are told apart syntactically rather than by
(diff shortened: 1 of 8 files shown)


devops_bench/agents/adk/agent.py
```diff
@@ -0,0 +1,579 @@
...
+            session = await runner.session_service.create_session(
+                app_name=_APP_NAME, user_id=_USER_ID
+            )
+            message = types.Content(role="user", parts=[types.Part(text=prompt)])
+            async for event in runner.run_async(
+                user_id=_USER_ID, session_id=session.id, new_message=message
+            ):
+                events.append(_dump_event(event))
+        finally:
+            await runner.close()
```
itssimrank: This should be wrapped with contextlib.suppress(asyncio.CancelledError) to protect the runner teardown from the timeout cancellation.

devops_bench/agents/adk/agent.py
```diff
@@ -0,0 +1,579 @@
...
+
+    path = pathlib.Path(os.path.expanduser(spec))
+    if not path.is_dir():
+        return _import_file(path, path.stem)
+
+    _ensure_on_sys_path(path.parent)
+    if (path / "__init__.py").is_file():
+        # A real package: import it by name so relative imports inside the
+        # agent resolve against the package rather than breaking.
+        return importlib.import_module(path.name)
```
itssimrank: This should check the cached module's file path actually lives under the requested directory and raise ConfigError on mismatch.

itssimrank (approved):

devops_bench/agents/adk/agent.py
jessie1111101: Agreed on the problem, but `contextlib.suppress` on its own doesn't actually fix it, so I went a bit further — pushed in 22d79ab.

I tried the suppress first and wrote a test for it, and the test passed with and without the change. Two reasons:

- The plain timeout path only cancels once. `wait_for` cancels the coroutine, the cancellation propagates into the `finally`, and `close()` then runs to completion on its own. Nothing to suppress.
- When there *is* a second cancellation (a Ctrl-C landing on a run whose budget already blew), suppressing the `CancelledError` doesn't make `close()` finish — it just swallows the error locally. The subprocess still leaks.

So teardown now runs as its own task and is awaited through `asyncio.shield`, with the wait retried rather than returning early, bounded by `_CLOSE_TIMEOUT_SEC = 30` so a wedged MCP server can't turn a reported timeout into a hang. That's `_close_quietly`.

Test is `test_close_quietly_finishes_under_repeated_cancellation` — it cancels the owner three times and asserts teardown reaches the end. It fails against a bare `await runner.close()` (stops at `started`) and passes with the fix.

devops_bench/agents/adk/agent.py
jessie1111101: Agreed, fixed in 22d79ab. `import_module` checks `sys.modules` before `sys.path`, so a stale `PYTHONPATH` entry or a name collision could silently load a different agent and still report success.

Added `_verify_imported_from`, which checks the resolved `__file__` lives under the requested directory and raises `ConfigError` naming both paths if not. Two tests cover it — the collision case and a genuine package, so the guard can't just reject everything.

Worth noting the non-package branch was never exposed: `_import_file` force-registers into `sys.modules` itself, so it always gets the file it asked for.

janetkuo (approved):
/lgtm

PR #148, opened 2026-09-01
feat(verification): add the identity_preserved verifier

jessie1111101:
## What this adds

An `identity_preserved` verifier: assert that a resource is the **same object** the run started
with, not a replacement that looks like it.

Needed by #105, which uses it five times.

## Why

`kubectl delete && kubectl apply` — or any equivalent "erase and rebuild" — produces a Deployment
that is byte-identical to the original in every spec and status field a naive check would grade,
including the image tag. Every remediation check we have passes on it.

This is not hypothetical. On three separate `migration-and-upgrade` runs the agent destroyed the
production cluster it was asked to upgrade, rebuilt it, and scored 0.7–0.8 with `success: True`.
Nothing in the pipeline could see it.

Kubernetes assigns a new server-generated `metadata.uid` on every create, and a client can never set
it. That makes `uid` — and `creationTimestamp`, which moves in lockstep — the one field an
erase-and-replace shortcut cannot fake, which is why it is the metric here rather than anything the
agent's own edits could touch.

## Design notes worth reviewing

- **The baseline travels with the object.** The fixture records the pre-run `uid` and
  `creationTimestamp` as annotations on the resource itself during setup, before the agent starts.
  Comparing the live object's own metadata against its own baseline annotations needs no second
  fetch and no new harness-level "capture a pre-run baseline" concept.
- **NotFound is a fail, not an error.** If the resource is gone the check returns `fail` with
  `"…no longer exists; it was deleted, not updated in place"`. Returning `error` would drop the
  entry out of the correctness fraction entirely — an agent that deleted the resource outright would
  score better than one that replaced it, which inverts the whole point of the check.

## Testing

10 new tests. Full suite green: 1320 passed. `ruff check` and `ruff format --check` clean.

## Dependencies

None. #74 has merged, so the stacked commit is gone and this is a single-commit PR against `main`.

/kind feature

```release-note
Add an `identity_preserved` verifier that detects a resource being deleted and recreated rather than updated in place, using the server-generated `metadata.uid` a client cannot fake.
```

### DIFF
--- a/devops_bench/verification/verifiers/__init__.py
+++ b/devops_bench/verification/verifiers/__init__.py
@@ -14,6 +14,9 @@
 
 """Concrete single-condition verifiers."""
 
+from devops_bench.verification.verifiers.identity_preserved import (
+    IdentityPreservedVerifier,
+)
 from devops_bench.verification.verifiers.pod_healthy import PodHealthyVerifier
 from devops_bench.verification.verifiers.resource_property import (
     ResourcePropertyVerifier,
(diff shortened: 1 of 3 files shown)


jessie1111101:
@coderabbitai review

PR #151, opened 2026-09-03
feat(metrics): zero the outcome when a run cheats

isadominguez314:
> **Builds on #139** (`feat/detection-core`). Its commits appear in this PR's
> diff because PRs cannot stack across a fork — the code is unchanged from #139.
> **Incremental diff (this PR's own work):**
> https://github.com/isadominguez314/devops-bench/compare/feat/detection-core...feat/integrity-gate

---

Detection has been flag-only: a run that read the benchmark's own material scored exactly as if it had not. Add an always-on, deterministic IntegrityMetric that turns a flagged cheating_report into a catastrophic zero, so the run stays on the leaderboard as a visible zero rather than disappearing from it the way validated=False would have done.

The gate emits IntegrityCatastrophic rather than reusing VerificationCatastrophic: the scores map is last-write-wins, so a clean integrity check sharing that key would erase a real task catastrophic. The two keys live together in core.score_keys and are read from there by both the pipeline and the normalizer, so the row's catastrophic flag cannot drift from the zero applied to outcomeScore.

A no_data report (errored run, or detection disabled) emits nothing -- having seen nothing is not innocence. The outcome finalizer no longer returns early on a gated run whose correctness sources all abstained, because a null outcomeScore drops the row out of leaderboard aggregates -- exactly the erasure a visible zero exists to prevent -- so outcomeScore is never left null once a gate has fired: the composite takes a synthesized correctness of 0.0 and the reason string reports c=n/a. Detection itself stays flag-only and never mutates a record; but its report is now consulted by scoring.

## Summary by CodeRabbit

- **New Features**
  - Added trajectory-based detection for potential access to sensitive benchmark material.
  - Reports now identify flagged, clean, or unavailable detection results for each record.
  - Added optional workspace inventory checks for artifacts from prior runs.
  - Added configurable detection rules and environment controls.
  - Added an integrity score that can mark affected runs as catastrophic and zero their outcome score.

- **Documentation**
  - Added guidance covering detection behavior, configuration, reports, metrics, and known limitations.
  - Clarified catastrophic outcome reporting across integrity and verification safeguards.


devops_bench/metrics/integrity.py
```diff
@@ -0,0 +1,151 @@
...
+
+* **Deterministic.** The verdict comes from the regex detector, never a judge.
+  Catastrophic safeguards hard gate the outcome, so they must be a
+  check tree — see the note on
+  :data:`~devops_bench.core.score_keys.JUDGED_RECOVERABLE_KEY`.
+* **Always-on.** No task opts in. Integrity is not a property a task declares;
+  it applies to every run of every task. One caveat the harness imposes rather
+  than this metric: scoring as a whole is skipped for ``status: "failed"``
+  records, so a cheat that also crashed is never gated (see the known
+  limitations in ``docs/components/detection.md``).
```
itssimrank: There's still a few instances that use detection.md instead of cheat-detection.md.

itssimrank (approved):

janetkuo (approved):
looks good but https://github.com/kubernetes-sigs/devops-bench/pull/139 needs fixing

janetkuo:
/hold 

until https://github.com/kubernetes-sigs/devops-bench/pull/139 is in

janetkuo (approved):
/lgtm

Issue #138, opened 2026-08-25
Support benchmarking ADK-built agents


jessie1111101:
## Goal

Benchmark agents built with the [Agent Development Kit](https://google.github.io/adk-docs/), so ADK teams can run their agents against devops-bench tasks and get a scored trajectory back. Pathfinder is the first consumer.

**Constraint that shapes everything below:** the Pathfinder team cannot run against open source. Changes have to be mirrored into google3 and run there, so every gap we miss costs a full mirror round trip. The plan is therefore to batch, not to iterate.

## Done

- [x] #137 — `feat(agents): add an ADK agent harness`. Registers `--agent-type adk`. Imports the agent named by `AGENT_TARGET`, deep-copies it, applies MCP/skills/rules to **every** agent in the tree, drives it through ADK's `Runner`, folds the event stream into a trajectory.
- [x] #136 — `fix(deps): cap mcp below 2.0`.

## PRs needed

Ordered. 1 is the only one that silently corrupts results.

### 1. Fix output extraction for remote agents

For an A2A agent, `AgentResult.output` comes back as the **last artifact's** text. The actual final answer sits in the task's `status.message` and is dropped. Confirmed with `use_legacy` both true and false, so it is consistent behaviour, not a legacy quirk.

Impact: the bench scores the agent against the wrong text. Any Pathfinder run before this lands is meaningless. Parser-only change, unit-testable from a recorded event.

### 2. Declare the `a2a` extra

Today the extra is bare `adk = ["google-adk>=2.7.1"]`. A2A needs `google-adk[a2a]` plus `a2a-sdk[grpc]` (grpcio + grpcio-tools). Packaging only.

### 3. Build the A2A client from config

So a caller gets a working gRPC connection from bench config instead of hand-writing a wrapper module — and, more importantly, without tripping the hazard in the "Validated" section below.

### 4. Sub-agent attribution in the trajectory

Two independent halves, both parser-only:

- **Native ADK trees** — ADK stamps `author` on every event and the parser discards it.
- **A2A agents** — per-sub-agent artifacts arrive in the event's `custom_metadata` and the parser ignores that field entirely.

This is FR2. See below for exactly how much of it comes free.

### 5. Smaller, unchanged from before

- Wire `AGENT_MAX_TURNS` to ADK's `RunConfig.max_llm_calls` (today the cap is ADK's default of 500).
[...]

PR #159, opened 2026-09-11
fix(agents): read a remote ADK agent's answer from the A2A envelope

jessie1111101:
## What

For a remote agent (`RemoteA2aAgent`), `AgentResult.output` was the **last artifact's** text rather than the agent's answer. The bench graded the wrong text — silently, with `errors: []` and `status: success`.

The answer lives in the A2A task envelope ADK attaches at `custom_metadata['a2a:response']`, as `status.message`. The `content.parts` ADK builds alongside it mirror the trailing artifact. The parser only read `content.parts`.

In the recorded event the two disagree:

| Field | Text |
| --- | --- |
| `status.message.parts[0].text` | `RCA: node memory pressure evicted the pod.` — the answer |
| `content.parts[0].text` | `node-1 mem = 0.97` — the trailing artifact |

Reproduced with `use_legacy` both true and false, so this is consistent behaviour rather than a legacy quirk.

## Change

Parser-only, in `devops_bench/agents/adk/parsing.py`. The task's state now decides what reaches `output`:

| Task state | What becomes the answer |
| --- | --- |
| `completed` | `status.message` — falling back to the event text when the task carries no message |
| `failed` / `canceled` / `rejected` | Nothing. The failure is recorded on `errors` instead |
| anything else (`working`, `input_required`, …) | The event's own text, as before |

`function_call` / `function_response` parts on the same event are still folded into the trajectory normally.

## Compatibility

A local agent's events carry no `a2a:response`, so nothing about them changes.

<details>
<summary><b>Why the state has to gate both halves of the event</b></summary>

**Only `completed` carries an answer.** ADK also emits streaming updates with a status message — `working` most often, and `input_required` / `auth_required` are no more final — and that text is progress commentary. Appending it would put narration ahead of the real answer in the output the judge grades.

**A failure needs both halves suppressed, not just the status message.** The message is a failure notice, and the `content.parts` mirror the trailing artifact — so letting the event text fall through as usual would reinstate the original bug for exactly the runs that have no answer at all. The record is still written as `status: "success"` and scored, so whatever reaches `output` is graded as the response.

**States are normalized before the lookup.** The proto types spell the enum `TASK_STATE_FAILED`; the pydantic ones spell it `failed`.

</details>

## Testing

`uv run pytest -q` — 1379 passed. Seven tests cover this path: the status message preferred over the artifact, each failure state contributing nothing (parametrized over all three), a completed task with no status message falling back to the event text, a lowercase state spelling, a `working` event left alone, and tool calls still folded on an A2A event. Four fail on `main`.

The fixture is shaped after a real `RemoteA2aAgent` run against an A2A gRPC server (`GrpcHandler` + `grpc.aio`, AgentCard advertising the GRPC binding), with the payload text replaced by neutral stand-ins.

`docs/components/agents.md` gains a "Remote agents" subsection: where the answer lives, and why a remote agent's trajectory is normally empty with token counts `None`.

`uv run ruff check devops_bench tests` and `ruff format --check` clean.

## Context

First of the follow-ups tracked in #138, and the only one that silently corrupts results. The rest — declaring the `a2a` extra (#171), building the A2A client from config (#172), and sub-agent attribution (#165) — are separate PRs.

## Summary by CodeRabbit

- **Bug Fixes**
  - Completed remote agent tasks now use the authoritative task status message when available.
  - Non-terminal updates use regular event content instead of being treated as final answers.
  - Failed, canceled, and rejected tasks are reported as errors, with their artifacts excluded from output.
  - Completed tasks fall back to event content when no status message is provided.
  - Tool-call details remain supported during remote agent event processing.

- **Documentation**
  - Added guidance on remote agent responses, task statuses, trajectories, and token reporting.

### DIFF
--- a/devops_bench/agents/adk/parsing.py
+++ b/devops_bench/agents/adk/parsing.py
@@ -30,6 +30,12 @@ Each event carries a ``content.parts`` list. Three part shapes matter:
 Calls and responses are correlated by the ``id`` ADK stamps on both sides, so a
 call and its result fold into one :class:`~devops_bench.agents.result.ToolCall`
 rather than two trajectory entries.
+
+An event from a remote A2A agent also carries the raw task envelope under
+``custom_metadata['a2a:response']``. That matters because the agent's actual
+answer lives in the task's ``status.message``, while the ``content.parts`` ADK
+builds alongside it hold a *mirror of the last artifact* — so the envelope, not
+the content, is what the judge should grade.
 """
 
 from __future__ import annotations
@@ -52,6 +58,15 @@ _USAGE_FIELDS: dict[str, str] = {
     "thoughts_token_count": "reasoning",
     "total_token_count": "total",
 }
+
+# Where ``RemoteA2aAgent`` stashes the raw task envelope on the event it builds.
+# ADK's converter does not export the key, so it is spelled out here.
+_A2A_RESPONSE_KEY = "a2a:response"
+
+# Terminal ``TaskState`` values meaning the remote agent never answered. The
+# proto types serialize the enum as ``TASK_STATE_FAILED`` and the pydantic types
+# as ``failed``, so states are normalized to the bare slug before the lookup.
+_A2A_FAILURE_STATES: frozenset[str] = frozenset({"failed", "canceled", "rejected"})
 
 
 def _int_or_none(value: object) -> int | None:
@@ -76,6 +91,45 @@ def _is_user_content(event: Mapping[str, Any]) -> bool:
     """
     content = event.get("content")
     return isinstance(content, Mapping) and content.get("role") == "user"
+
+
+def _a2a_task(event: Mapping[str, Any]) -> Mapping[str, Any] | None:
+    """Return the A2A task envelope ``RemoteA2aAgent`` attached, if any."""
+    metadata = event.get("custom_metadata")
+    if not isinstance(metadata, Mapping):
+        return None
+    response = metadata.get(_A2A_RESPONSE_KEY)
+    return response if isinstance(response, Mapping) else None
+
+
+def _a2a_status(task: Mapping[str, Any]) -> tuple[str | None, str | None]:
+    """Pull the reported state and final message text out of a task envelope.
+
+    Returns:
+        A ``(state, text)`` tuple. ``state`` is normalized to the bare slug
+        (``"completed"``, not ``"TASK_STATE_COMPLETED"``). ``text`` is ``None``
+        when the task carried no status message, which is the case while a task
+        is still working.
+    """
+    status = task.get("status")
+    if not isinstance(status, Mapping):
+        return None, None
+
+    raw_state = status.get("state")
+    state = raw_state.lower().removeprefix("task_state_") if isinstance(raw_state, str) else None
+
+    message = status.get("message")
+    parts = message.get("parts") if isinstance(message, Mapping) else None
+    texts = (
+        [
+            part["text"]
+            for part in parts
+            if isinstance(part, Mapping) and isinstance(part.get("text"), str)
+        ]
+        if isinstance(parts, list)
+        else []
+    )
+    return state, ("".join(texts) or None)
 
 
 def _response_text(response: Any) -> tuple[str, bool]:
@@ -172,6 +226,10 @@ def parse_event_stream(
     the output text: the former would duplicate the text they are chunks of, and
     the latter is reasoning the judge should not grade as an answer.
 
+    An event carrying an A2A task envelope is read from the envelope instead: its
+    ``status.message`` is the remote agent's answer, and a terminal state other
+    than completed lands on ``errors``.
+
     Args:
         events: Serialized ``Event`` mappings in the order ADK yielded them.
 
@@ -207,6 +265,17 @@ def parse_event_stream(
         partial = bool(event.get("partial"))
         user_content = _is_user_content(event)
 
+        # A remote agent's answer is the A2A task's status message. Take it in
+        # place of the event's own text, which mirrors the trailing artifact.
+        a2a_text: str | None = None
+        task = _a2a_task(event)
+        if task is not None:
+            state, a2a_text = _a2a_status(task)
+            if state in _A2A_FAILURE_STATES:
+                errors.append(f"event {index}: remote A2A task {state}")
+            if a2a_text and not partial:
+                output_parts.append(a2a_text)
+
         for part in _parts(event):
             if not isinstance(part, Mapping):
                 continue
@@ -239,6 +308,7 @@ def parse_event_stream(
                 and not partial
                 and not user_content
                 and not part.get("thought")
+                and a2a_text is None
             ):
                 output_parts.append(text)
 
--- a/docs/components/agents.md
+++ b/docs/components/agents.md
@@ -213,6 +213,20 @@ persisted.
 > means `transfer_to_agent` hops and the delegate's own calls are flattened
 > together.
 
+### Remote agents
+
+A `RemoteA2aAgent` is a `BaseAgent`, so the harness drives one unmodified. What
+differs is where its answer lives. ADK attaches the raw A2A task envelope to the
+event under `custom_metadata['a2a:response']`, and the answer is that task's
+`status.message` — the `content.parts` built alongside it mirror the trailing
+artifact instead. The parser reads the envelope, and records a terminal task
+state other than completed (`failed`, `canceled`, `rejected`) on the result's
+errors, so a remote failure is not scored as an answer.
+
+The trajectory of a remote agent is empty and its token counts are `None`: the
+tool calls and LLM calls happen on the far side of the boundary and are never
+reported as ADK parts.
+
 The agent runs with the harness-owned workspace as the process working
 directory, matching the `cwd` the CLI harnesses hand their subprocess. An agent
 with filesystem tools that writes a relative path (`report.md`) therefore lands
--- a/tests/unit/agents/test_agents_adk.py
+++ b/tests/unit/agents/test_agents_adk.py
@@ -143,6 +143,52 @@ MCP_CALL_EVENT = {
     "id": "bb22",
 }
 
+# Recorded from a ``RemoteA2aAgent`` driven against a real A2A gRPC server. Note
+# that ``content.parts`` mirrors the *trailing artifact* while the answer sits in
+# ``status.message`` — the two disagree, which is the point of the fixture.
+A2A_EVENT = {
+    "content": {"parts": [{"text": "monarch node.mem = 0.97"}], "role": "model"},
+    "custom_metadata": {
+        "a2a:task_id": "494405c7-b9be-441f-beae-be07ba49b5b7",
+        "a2a:context_id": "64c847ad-f70e-42f2-b24e-2b88e3550780",
+        "a2a:request": {
+            "messageId": "6bbd7789-6c50-49bf-a713-6f83a37f4f58",
+            "role": "ROLE_USER",
+            "parts": [{"text": "Diagnose b/123", "metadata": {"is_user_input": True}}],
+        },
+        "a2a:response": {
+            "id": "494405c7-b9be-441f-beae-be07ba49b5b7",
+            "contextId": "64c847ad-f70e-42f2-b24e-2b88e3550780",
+            "status": {
+                "state": "TASK_STATE_COMPLETED",
+                "message": {
+                    "messageId": "587e9b357eb547629ef6d675a01d60b5",
+                    "role": "ROLE_AGENT",
+                    "parts": [{"text": "RCA: node memory pressure evicted the pod."}],
+                },
+                "timestamp": "2026-09-11T16:47:45.085206Z",
+            },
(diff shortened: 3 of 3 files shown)


devops_bench/agents/adk/parsing.py
```diff
@@ -207,6 +265,17 @@ def parse_event_stream(
...
+        # A remote agent's answer is the A2A task's status message. Take it in
+        # place of the event's own text, which mirrors the trailing artifact.
+        a2a_text: str | None = None
+        task = _a2a_task(event)
+        if task is not None:
+            state, a2a_text = _a2a_status(task)
+            if state in _A2A_FAILURE_STATES:
+                errors.append(f"event {index}: remote A2A task {state}")
+            if a2a_text and not partial:
+                output_parts.append(a2a_text)
```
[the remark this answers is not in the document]
jessie1111101: Good catch, fixed in the push above.

I went a little broader than the suggested diff. `working` isn't the only non-final state that carries a status message — `input_required` and `auth_required` are equally non-final, so gating on `state == "completed"` would still let those two through. The check is now a terminal-state set:

```python
_A2A_FINAL_STATES: frozenset[str] = _A2A_FAILURE_STATES | {"completed"}
```

There's also a second effect worth naming, since it changes the fix: the event's own text is gated on `a2a_text is None`, so a non-final status message didn't only put narration ahead of the answer — it *suppressed* the update's own content in favour of the progress note. Leaving `a2a_text` unset on a non-terminal state restores both.

I also dropped a line from `_a2a_status`'s docstring claiming a missing status message "is the case while a task is still working" — the same evidence disproves it.

Regressions added as requested: `working` → `completed` asserting only the final text is returned, plus a parametrized case over `TASK_STATE_INPUT_REQUIRED` / `TASK_STATE_AUTH_REQUIRED`. Confirmed all three fail without the parser change.

jessie1111101:
@coderabbitai review

jessie1111101:
@coderabbitai rate limit

devops_bench/agents/adk/parsing.py
```diff
@@ -207,6 +276,23 @@ def parse_event_stream(
         partial = bool(event.get("partial"))
         user_content = _is_user_content(event)
 
+        # A remote agent's answer is the A2A task's status message. Take it in
+        # place of the event's own text, which mirrors the trailing artifact.
+        a2a_text: str | None = None
+        task = _a2a_task(event)
+        if task is not None:
+            state, status_text = _a2a_status(task)
+            if state in _A2A_FAILURE_STATES:
```
itssimrank, line 286: Does this also cover a terminal state whose status carries no message? Could you add a test case for it as well?

itssimrank (approved):

jessie1111101:
@coderabbitai review

devops_bench/agents/adk/parsing.py
```diff
@@ -207,6 +276,23 @@ def parse_event_stream(
...
+                errors.append(f"event {index}: remote A2A task {state}")
+            # Only a terminal state's message is the answer. Leaving ``a2a_text``
+            # unset on a streaming update also lets that event's own text fall
+            # through as before, rather than being suppressed in favour of the
+            # progress note that displaced it.
+            if state in _A2A_FINAL_STATES:
+                a2a_text = status_text
+            if a2a_text and not partial:
+                output_parts.append(a2a_text)
+
```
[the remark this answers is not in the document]
jessie1111101: Confirmed and fixed in the push above — though not quite as proposed, because restricting the status message to `completed` on its own reintroduces the bug this PR opened with.

The chain you traced is exactly right: `_build_success_record` sets `status: "success"` regardless of `errors`, `_score` skips only `"failed"` records, and `_build_context` hands `output` to the judge. A failure notice was being graded as the answer.

The catch is the text gate further down: the event's own parts are appended when `a2a_text is None`. Leaving `a2a_text` unset for a failure state therefore falls back to `content.parts` — the trailing-artifact mirror this whole path exists to keep out of graded output. Same misattribution, just narrowed to failures.

So a failed task now contributes nothing either way, via a separate suppression flag:

```python
if state in _A2A_FAILURE_STATES:
    a2a_failed = True
    errors.append(f"event {index}: remote A2A task {state}")
elif state == _A2A_ANSWER_STATE:
    a2a_text = status_text
```

with `and not a2a_failed` added to the text gate. `_A2A_FINAL_STATES` is retired in favour of `_A2A_ANSWER_STATE = "completed"`, since "final" was conflating the two meanings. Covered by a parametrized regression across all three failure states asserting empty output when the envelope carries only an artifact.

devops_bench/agents/adk/parsing.py
jessie1111101: Good catch — it did, but only by accident, and it wasn't tested. Added `test_parse_event_stream_falls_back_to_content_on_a_completed_task_with_no_message`.

The behaviour I settled on splits on whether the task actually produced anything:

- **`completed`, no status message** → falls back to the event's own text, same as a non-terminal state. The task did produce something, and the artifact is the only text there is.
- **`failed` / `canceled` / `rejected`, message or not** → contributes nothing to output. There is no answer to fall back to, and the artifact would just pose as one.

That second half also covers the Major finding the bot raised in parallel, so the two questions turned out to be the same flag. Both paths are now covered — the completed-without-message fallback, and a parametrized case across all three failure states.

jessie1111101:
@coderabbitai full review

jessie1111101:
@coderabbitai full review

pradeepvrd (approved):

pradeepvrd:
LGTM. Thanks for the fix

jessie1111101:
@coderabbitai review

devops_bench/agents/adk/parsing.py
```diff
@@ -53,6 +59,23 @@
     "total_token_count": "total",
 }
 
+# Where ``RemoteA2aAgent`` stashes the raw task envelope on the event it builds.
+# ADK's converter does not export the key, so it is spelled out here.
+_A2A_RESPONSE_KEY = "a2a:response"
```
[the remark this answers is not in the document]
jessie1111101: Partially — the two module constants are annotated in the push above, the rest I'd rather not.

The inconsistency is real: `_A2A_FAILURE_STATES: frozenset[str]` sits three lines away from two bare constants. Both now carry `: str`.

Not `Final[str]` though — `Final[` appears nowhere in this repo, so it'd be a new construct introduced by one PR rather than a convention followed. Plain annotations match what's already here.

On `A2A_EVENT`: its three neighbours in the same file (`MCP_CALL_EVENT`, `MCP_RESPONSE_EVENT`, `CALL_EVENT`) are all unannotated module-level fixtures. Annotating just the one this PR happens to touch would make it the odd one out. Happy to annotate all of them if you'd like, but that's a separate sweep, not this fix.

docs/components/agents.md
```diff
@@ -213,6 +213,27 @@ persisted.
...
+
+A failed task (`failed`, `canceled`, `rejected`) is recorded on the result's
+errors and contributes **nothing** to the output: neither its status message,
+which is a failure notice, nor its content parts, which mirror the artifact. The
+record is still written as `status: "success"` and scored, so either one left in
+the output would be graded as the agent's answer.
+
+The trajectory of a remote agent is empty and its token counts are `None`: the
+tool calls and LLM calls happen on the far side of the boundary and are never
+reported as ADK parts.
```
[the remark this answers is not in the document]
jessie1111101: Agreed, the claim was too strong — and this PR's own `test_parse_event_stream_still_folds_tool_calls_on_an_a2a_event` contradicts it. `_accumulate_usage` also runs on every event, A2A or not. Reworded:

> A remote agent's trajectory is normally empty and its token counts `None` […] This is a property of what the remote reports, not a rule the parser enforces — an A2A event carrying `function_call` / `function_response` parts is folded into the trajectory like any other, and `usage_metadata` is accumulated wherever ADK supplies it.

devops_bench/agents/adk/parsing.py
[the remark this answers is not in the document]
jessie1111101: Fair enough — you're right that a pre-existing gap doesn't exempt new code. Annotated in the push above as `A2A_EVENT: dict[str, Any]`, with `Any` added to the imports.

Leaving the three neighbouring fixtures for a separate cleanup, as you suggested, so this PR's diff stays about the A2A output bug.

jessie1111101:
@coderabbitai full review

devops_bench/agents/adk/parsing.py
```diff
@@ -53,6 +59,23 @@
...
 }
 
+# Where ``RemoteA2aAgent`` stashes the raw task envelope on the event it builds.
+# ADK's converter does not export the key, so it is spelled out here.
+_A2A_RESPONSE_KEY: str = "a2a:response"
+
+# Terminal ``TaskState`` values meaning the remote agent never answered. The
+# proto types serialize the enum as ``TASK_STATE_FAILED`` and the pydantic types
+# as ``failed``, so states are normalized to the bare slug before the lookup.
+_A2A_FAILURE_STATES: frozenset[str] = frozenset({"failed", "canceled", "rejected"})
```
janetkuo: [Non-blocking] In A2A v0.3 protobuf/gRPC definitions, the cancelled enum is serialized as `"TASK_STATE_CANCELLED"` (with double 'l'). After stripping the prefix, `state` normalizes to `"cancelled"`, which does not match `_A2A_FAILURE_STATES`. Add `"cancelled"` to `_A2A_FAILURE_STATES` so both v0.3 and v1.0 specifications are handled.

devops_bench/agents/adk/parsing.py
```diff
@@ -207,6 +276,32 @@ def parse_event_stream(
...
+            if state in _A2A_FAILURE_STATES:
+                # A failed task has no answer, and neither half of the event is
+                # a stand-in for one: the status message is a failure notice,
+                # and the content parts mirror the trailing artifact. The record
+                # is still written as ``status: "success"`` and scored, so
+                # whichever one reaches ``output`` gets graded as the response.
+                # Report the failure and contribute nothing.
+                a2a_failed = True
+                errors.append(f"event {index}: remote A2A task {state}")
+            elif state == _A2A_ANSWER_STATE:
```
janetkuo: [Blocker] When a remote task emits non-terminal updates (e.g. state is `"working"`, `"input_required"`, or `"auth_required"`), `a2a_text` remains `None` and `a2a_failed` is `False`. In the part loop below (lines 337-338), `and a2a_text is None and not a2a_failed` evaluates to `True`, causing the parser to fall back to appending `content.parts` to `output_parts`.

This creates two issues:
1. Multi-event streaming runs: Any intermediate artifacts or progress text emitted during `"working"` states leak into `output_parts` and get prepended to the final answer (as observed in `test_parse_event_stream_ignores_a_working_tasks_status_message`).
2. Uncompleted runs: For runs that end in `"input_required"` or `"auth_required"` without answering, the parser extracts the trailing artifact from `content.parts` and returns it as output with `errors = []` (as asserted in `test_parse_event_stream_ignores_a_non_terminal_status_message`). This reintroduces the silent grading of artifacts for runs that produced no answer.

Non-terminal A2A task events must not fall through to appending `content.parts` to `output_parts`. For `"input_required"` and `"auth_required"`, the run did not complete and should be recorded on `errors` rather than returning trailing artifacts.

docs/components/agents.md
```diff
@@ -213,6 +213,30 @@ persisted.
...
+### Remote agents
+
+A `RemoteA2aAgent` is a `BaseAgent`, so the harness drives one unmodified. What
+differs is where its answer lives. ADK attaches the raw A2A task envelope to the
+event under `custom_metadata['a2a:response']`, and the answer is that task's
+`status.message` — the `content.parts` built alongside it mirror the trailing
+artifact instead. The parser reads the envelope, and takes the status message as
+the answer only on `completed` — a `working` / `input_required` /
+`auth_required` message is progress commentary, and the event's own text is used
+instead.
```
janetkuo: [Question] The text states that for `working` / `input_required` / `auth_required`, "the event's own text is used instead". Since the event's own text mirrors the artifact, using it during these states reintroduces artifact leakage into the graded output. Once non-terminal event handling in `parsing.py` is updated, please update this description accordingly.

tests/unit/agents/test_agents_adk.py
```diff
@@ -143,6 +144,53 @@
...
+                    "role": "ROLE_AGENT",
+                    "parts": [{"text": "RCA: node memory pressure evicted the pod."}],
+                },
+                "timestamp": "2026-09-11T16:47:45.085206Z",
+            },
+            "artifacts": [
+                {
+                    "artifactId": "5d57761d-7b73-4100-8457-d26419e3b0a8",
+                    "name": "triage_agent",
+                    "parts": [{"text": "matched skill gke-node-pressure"}],
```
janetkuo: [Nit] `"gke-node-pressure"` retains a provider-specific prefix in a generic harness test fixture. Per `AGENTS.md` vendor-neutrality rules, replace this with a neutral identifier such as `"matched skill node-pressure"`.

janetkuo (approved):
/lgtm

PR #160, opened 2026-09-11
fix(agents): send agy a model id it accepts

isadominguez314:
Every `agy` run dies at startup with `invalid model selection`. This is **not sandbox-specific** — an ambient, un-sandboxed run fails identically, and has since the harness was written. It stayed latent because agy had no exercised arm to fail in.

> Staged on the `pradeepvrd/devops-bench` integration fork as pradeepvrd/devops-bench#28. Independent of the sandbox stack and of the image PR #161.

## The incident

Sandbox validation run on the bastion, 2026-09-10 ~16:19Z, agy 1.2.0, sandboxed GKE run:

```
error: invalid model selection (--model "gemini-3.1-pro-preview" --effort ""):
model gemini-3.1-pro-preview is not recognized as a known model or custom model in settings
```

Exit 1 in roughly two seconds, no conversation DB, no transcript. That arm was then validated under the display-name spelling `AGENT_MODEL="Gemini 3.1 Pro (Low)"` as a workaround, which is why that run's model label does not match the rest of the matrix — and why it is evidence for the sandbox seam, not a comparable agent-quality result.

## Root cause

`AGENT_MODEL` is a single global value shared by every arm and by the judge (`scripts/bastion/vm-setup.sh:209` = `gemini-3.1-pro-preview`), spelled for Vertex, which *requires* the `-preview` suffix.

`_resolve_model_name` only stripped a provider prefix, so the Vertex id reached agy's command line verbatim.

## What agy 1.2.0 actually accepts

No `-preview` suffix, and a reasoning tier named **exactly once** — as `--effort`, a slug suffix, or a display-name parenthetical.

| spelling | result |
| --- | --- |
| `gemini-3.1-pro-preview` (what the matrix sends) | rejected — unrecognized |
| `gemini-3.1-pro-preview --effort=low` | rejected — `--effort is not supported` |
| `gemini-3.1-pro` | rejected — `requires --effort (available: low, high)` |
| `gemini-3.1-pro --effort=low` | **ok** — canonical form |
| `gemini-3.8-flash --effort=medium` | **ok** |
| `gemini-3.1-pro-low` | **ok** |
| `Gemini 3.1 Pro (Low)` | **ok**, case-insensitive |
| `Gemini 3.1 Pro (Low) --effort=low` | rejected — collides |

The catalogue is 11 selections: `Gemini 3.8/3.7/3.6 Flash (High|Medium|Low)` and `Gemini 3.1 Pro (High|Low)`. Everything is Flash except the two Pro entries. Any `gemini-2.5-*` id is rejected outright.

## The fix

Normalise inside the antigravity harness. Respelling `AGENT_MODEL` for the agy arm instead would desynchronise that arm's label from every other arm in the same matrix, and Vertex needs exactly the suffix agy rejects — so the quirk belongs in the one harness that has it.

`google/gemini-3.1-pro-preview` → `--model gemini-3.1-pro --effort high`, logged when the rewrite happens.

Emitting the bare slug plus `--effort`, rather than reassembling `gemini-3.1-pro-low`, is the one form that cannot collide. No lookup table, so nothing to maintain as the catalogue grows — agy validates the result and still fails loud on anything it does not know.

A tier already spelled into the id is honoured and nothing is added, so the display-name workaround above keeps working.

### The tier is a scoring variable

agy has no untiered form, so the harness must supply one when `AGENT_MODEL` carries none — which is always, today. `low` versus `high` is a materially different agent, so choosing quietly would make the agy arm a different experiment than the arm it is compared against.

Default `high`, since the other harnesses run their model with no reasoning throttle and anything lower would hand agy a handicap that reads as a capability gap. Overridable via `AGENT_MODEL_EFFORT`, with an unknown tier rejected on the spot rather than passed through to fail later as agy's own error mid-arm — reject at load, not at use.

### settings.json now agrees with the flag

`modelConfigs.defaultModel` carries the same resolved spelling as `--model`.

Worth stating plainly, because it is a trap: **agy does not validate `defaultModel`.** An unknown value there is ignored in silence and the run falls back to another model — verified, a garbage value runs happily and self-reports as Gemini 3.8 Flash. The `--model` flag is the only validation there is, so it should not be guarding a value that settings quietly contradicts, and it should not be dropped as redundant later.

### GEMINI_MODEL removed

`_build_env` set it; **agy ignores it entirely**. Verified: a garbage value raises no error, and a valid `gemini-3.1-pro-low` does not change the model the run reports — identical to the no-env control. Copy-paste from the gemini_cli harness. It looked like a lever and was not.

## Verification

Unit suite green on this branch (1364 passed), ruff check and format clean. New coverage over the resolver table above, including the collision case, the display-name pass-through, and load-time rejection of an unknown `AGENT_MODEL_EFFORT`.

Every spelling in the table was exercised against agy 1.2.0 on the bastion, in the pinned image from #161, under the sandbox executor's real invocation (`--user <uid>:<gid>`, `HOME` on the workspace mount, metadata endpoint unreachable).

## Not in this PR

The `_resolve_binary` host-path leak into container argv — that belongs with the sandbox boundary that creates it, and is held in the harness-wiring PR staged as pradeepvrd/devops-bench#14.

## Summary by CodeRabbit

* **New Features**
  * Antigravity now supports configurable model reasoning effort through `AGENT_MODEL_EFFORT`.
  * Provider-qualified and preview model IDs are normalized automatically for compatible CLI usage.
  * Explicit model tiers are preserved, with a default effort level applied when none is specified.
  * Invalid effort values are rejected with a configuration error.

* **Documentation**
  * Added guidance on supported effort values, validation behavior, and model ID rewriting.

### DIFF
--- a/devops_bench/agents/cli/antigravity/agent.py
+++ b/devops_bench/agents/cli/antigravity/agent.py
@@ -19,6 +19,7 @@ from __future__ import annotations
 import json
 import os
 import pathlib
+import re
 import shutil
 import time
 from typing import TYPE_CHECKING
@@ -77,12 +78,81 @@ def _read_db_tokens(db_path: pathlib.Path) -> dict | None:
     return None
 
 
-def _resolve_model_name(model: str) -> str:
-    """Resolve a provider-qualified model id to the bare name ``agy`` expects.
+# agy names the reasoning tier separately from the model: every selection in its
+# catalogue is a base model plus one of these, and there is no untiered form --
+# a bare slug is refused with "requires --effort".
+_AGY_EFFORT_TIERS = ("low", "medium", "high")
 
-    e.g. ``"google/gemini-3.5-flash"`` -> ``"gemini-3.5-flash"``.
+# Vertex publishes preview model ids with this suffix. agy's catalogue does not
+# carry it and rejects the suffixed id outright, with or without --effort.
+_VERTEX_PREVIEW_SUFFIX = "-preview"
+
+# A display-name selection ("Gemini 3.1 Pro (Low)") already names its tier, and
+# agy errors if --effort is passed alongside one. Recognize it so it survives
+# untouched: it is the spelling ``agy --help`` steers operators towards.
+_DISPLAY_TIER_RE = re.compile(r"\((?:low|medium|high)\)\s*$", re.IGNORECASE)
+
+# The tier is a scoring variable, not a formatting detail -- `low` and `high`
+# are materially different agents. `high` is the least surprising default
+# because the other harnesses run their model with no reasoning throttle, so
+# anything lower would hand the agy arm a handicap that reads as a capability
+# gap in the results rather than as the configuration choice it is.
+_DEFAULT_AGY_EFFORT = "high"
+_EFFORT_ENV = "AGENT_MODEL_EFFORT"
+
+
+def _default_effort() -> str:
+    """Resolve the reasoning tier to request when the model id names none.
+
+    Returns:
+        The tier from ``AGENT_MODEL_EFFORT``, or ``high``.
+
+    Raises:
+        core.ConfigError: If ``AGENT_MODEL_EFFORT`` names an unknown tier.
+            Rejected here rather than passed through, so a typo fails on the
+            misconfiguration itself instead of surfacing seconds later as agy's
+            own startup error in the middle of a scored arm.
     """
-    return model.split("/")[-1]
+    effort = os.environ.get(_EFFORT_ENV, "").strip()
+    if not effort:
+        return _DEFAULT_AGY_EFFORT
+    if effort.lower() not in _AGY_EFFORT_TIERS:
+        raise core.ConfigError(
+            f"{_EFFORT_ENV}={effort!r} is not a reasoning tier agy accepts "
+            f"(known: {', '.join(_AGY_EFFORT_TIERS)})"
+        )
+    return effort.lower()
+
+
+def _resolve_model_name(model: str) -> tuple[str, str | None]:
+    """Resolve a matrix model id to the ``(model, effort)`` pair ``agy`` expects.
+
+    ``AGENT_MODEL`` is one value shared by every arm and by the judge, and it is
+    spelled for Vertex -- ``google/gemini-3.1-pro-preview``. agy accepts neither
+    the provider prefix nor the ``-preview`` suffix, and refuses any selection
+    that does not name a reasoning tier exactly once. Normalizing here confines
+    the quirk to the one harness that has it; respelling ``AGENT_MODEL`` instead
+    would desynchronize the agy arm's label from every other arm in the matrix.
+
+    e.g. ``"google/gemini-3.1-pro-preview"`` -> ``("gemini-3.1-pro", "high")``.
+
+    Args:
+        model: The configured model id, optionally provider-qualified.
+
+    Returns:
+        The id to pass as ``--model``, and the tier to pass as ``--effort`` --
+        or ``None`` for the tier when the id already names its own, in which
+        case ``--effort`` must be omitted or agy rejects the pair.
+    """
+    name = model.split("/")[-1].strip()
+    if _DISPLAY_TIER_RE.search(name):
+        return name, None
+    if name.endswith(_VERTEX_PREVIEW_SUFFIX):
+        name = name[: -len(_VERTEX_PREVIEW_SUFFIX)]
+    for tier in _AGY_EFFORT_TIERS:
+        if name.lower().endswith(f"-{tier}"):
+            return name[: -len(tier) - 1], tier
+    return name, _default_effort()
 
 
 def _build_settings(
@@ -93,7 +163,16 @@ def _build_settings(
     *,
     skills_enabled: bool = False,
 ) -> dict:
-    """Assemble the Antigravity ``settings.json`` payload for a run."""
+    """Assemble the Antigravity ``settings.json`` payload for a run.
+
+    ``model`` must already be the resolved spelling from
+    ``_resolve_model_name``, identical to the one passed as ``--model``. agy
+    does not validate ``defaultModel`` -- an unknown value there is ignored in
+    silence and the run falls back to a default model -- so the flag's loud
+    validation is the only guard, and it only guards a value settings agrees
+    with. The tier is deliberately not written here: it rides on ``--effort``,
+    and agy exposes no verified settings key for it.
+    """
     settings: dict = {}
     servers = cli_capabilities.build_mcp_servers(mcp_servers)
     if servers:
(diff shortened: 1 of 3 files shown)


devops_bench/agents/cli/antigravity/agent.py
```diff
@@ -77,12 +78,81 @@ def _read_db_tokens(db_path: pathlib.Path) -> dict | None:
     return None
 
 
-def _resolve_model_name(model: str) -> str:
-    """Resolve a provider-qualified model id to the bare name ``agy`` expects.
+# agy names the reasoning tier separately from the model: every selection in its
+# catalogue is a base model plus one of these, and there is no untiered form --
+# a bare slug is refused with "requires --effort".
+_AGY_EFFORT_TIERS = ("low", "medium", "high")
```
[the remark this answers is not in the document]
isadominguez314: Done in d3b679c — the five constants now carry explicit annotations.

tests/unit/agents/test_agents_cli_antigravity.py
```diff
@@ -334,17 +336,74 @@ def test_build_env_sets_auth_and_presets():
...
     assert env["GEMINI_API_KEY"] == "secret-key"
     assert env["GOOGLE_API_KEY"] == "secret-key"
-    assert env["GEMINI_MODEL"] == "gemini-3.5-flash"
     assert env["OTEL_SDK_DISABLED"] == "true"
 
 
-def test_build_env_resolves_provider_qualified_model_name():
-    # GEMINI_MODEL must match the bare id used by --model= and modelConfigs,
-    # not the raw "provider/model" form.
+def test_build_env_does_not_set_gemini_model():
```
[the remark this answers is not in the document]
isadominguez314: Done in d3b679c — `-> None` on the new tests and `pytest.MonkeyPatch` on the monkeypatch params. Only this PR's additions are annotated; the pre-existing unannotated tests in this file are out of scope here.

devops_bench/agents/cli/antigravity/agent.py
```diff
@@ -77,12 +78,81 @@ def _read_db_tokens(db_path: pathlib.Path) -> dict | None:
...
+        The tier from ``AGENT_MODEL_EFFORT``, or ``high``.
+
+    Raises:
+        core.ConfigError: If ``AGENT_MODEL_EFFORT`` names an unknown tier.
+            Rejected here rather than passed through, so a typo fails on the
+            misconfiguration itself instead of surfacing seconds later as agy's
+            own startup error in the middle of a scored arm.
     """
-    return model.split("/")[-1]
+    effort = os.environ.get(_EFFORT_ENV, "").strip()
```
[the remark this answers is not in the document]
isadominguez314: Declining this one deliberately. `AgentConfig` is provider-agnostic by design — every field on it is consumed by all harnesses, and `AGENT_MODEL_EFFORT` is an agy-only operator escape hatch (no other CLI names a reasoning tier separately from the model). The production path reads process env because that is the only way the matrix sets it, and the validation happens at first use so a typo still fails on the misconfiguration itself. If a second harness ever grows a tier knob, promoting it into the config seam is the right move; a single-consumer field on the shared config isn't.

isadominguez314:
@coderabbitai review

devops_bench/agents/cli/antigravity/agent.py
```diff
@@ -77,12 +78,81 @@ def _read_db_tokens(db_path: pathlib.Path) -> dict | None:
...
+    if not effort:
+        return _DEFAULT_AGY_EFFORT
+    if effort.lower() not in _AGY_EFFORT_TIERS:
```
eugeneng04: this isn't an issue currently, but for gemini 3.1 pro, it only has low and high, so "medium" will pass, even though it doesn't exist. So it will error out.

Maybe more docstrings would be useful? or raise something since theres already a lot of docstrings.

docs/components/agents.md
eugeneng04: This still says antigravity maps the model onto GEMINI_MODEL, but this PR removes that and test_build_env_does_not_set_gemini_model asserts it's gone. I think this should point at --model/--effort and the new "agy model ids" section.

devops_bench/agents/cli/antigravity/agent.py
```diff
@@ -77,12 +78,81 @@ def _read_db_tokens(db_path: pathlib.Path) -> dict | None:
...
+
+    Returns:
+        The id to pass as ``--model``, and the tier to pass as ``--effort`` --
+        or ``None`` for the tier when the id already names its own, in which
+        case ``--effort`` must be omitted or agy rejects the pair.
+    """
+    name = model.split("/")[-1].strip()
+    if _DISPLAY_TIER_RE.search(name):
+        return name, None
+    if name.endswith(_VERTEX_PREVIEW_SUFFIX):
```
eugeneng04: Since -preview is only stripped when it's at the end, google/gemini-3.1-pro-preview-low resolves to ("gemini-3.1-pro-preview", "low") and agy rejects it. Maybe split off the tier first and then strip -preview?

devops_bench/agents/cli/antigravity/agent.py
```diff
@@ -206,6 +288,21 @@ def _execute(
...
         binary = self._resolve_binary()
 
+        # Resolved once so the --model flag and settings.json cannot disagree,
+        # and before any workspace is built so a bad AGENT_MODEL_EFFORT fails
+        # here rather than after the run has started costing something.
+        model_name: str | None = None
+        effort: str | None = None
+        if self.config.model:
+            model_name, effort = _resolve_model_name(self.config.model)
+            if model_name != self.config.model:
```
eugeneng04: A name like `gemini-3.8-flash` doesn't change here, so --effort high gets added with nothing logged, even though the docs say "the rewrite is logged". Since the tier changes the score, we might want to log whenever effort is added.

tests/unit/agents/test_agents_cli_antigravity.py
eugeneng04: None of the tests check that --model/--effort actually end up in argv. Removing both argv.append calls still passes all 40 tests. Maybe assert --model=gemini-3.5-flash and --effort=high here?

devops_bench/agents/cli/antigravity/agent.py
```diff
@@ -77,12 +78,81 @@ def _read_db_tokens(db_path: pathlib.Path) -> dict | None:
...
 
-    e.g. ``"google/gemini-3.5-flash"`` -> ``"gemini-3.5-flash"``.
+# Vertex publishes preview model ids with this suffix. agy's catalogue does not
+# carry it and rejects the suffixed id outright, with or without --effort.
+_VERTEX_PREVIEW_SUFFIX: str = "-preview"
+
+# A display-name selection ("Gemini 3.1 Pro (Low)") already names its tier, and
+# agy errors if --effort is passed alongside one. Recognize it so it survives
+# untouched: it is the spelling ``agy --help`` steers operators towards.
+_DISPLAY_TIER_RE: re.Pattern[str] = re.compile(r"\((?:low|medium|high)\)\s*$", re.IGNORECASE)
```
eugeneng04: The tiers are hardcoded again here, so adding one to `_AGY_EFFORT_TIERS` (like xhigh) without updating this would send a display name with that tier a second --effort, which agy rejects. Maybe reference `_AGY_EFFORT_TIERS`?

eugeneng04 (approved):
/lgtm I think theres small fixes that are needed for future models/changes overall looks good

PR #161, opened 2026-09-11
build(docker): add the Antigravity CLI sandbox image

isadominguez314:
Adds `docker/Dockerfile.antigravity`. One file, no code changes.

> Staged on the `pradeepvrd/devops-bench` integration fork as pradeepvrd/devops-bench#27, where the sandbox seam it serves is already merged. The upstream PRs for that seam follow separately; this image is inert until they land, but it is independently reviewable and the pin is worth freezing now.

The sandbox seam documents an image contract — ship `docker/Dockerfile.<name>` carrying the shared tool floor with the CLI version-pinned and installed under **the exact binary name the harness invokes** — but no antigravity image existed, so `agy` had no sandboxed arm at all.

Deliberately not stacked on the harness-wiring PR (staged as pradeepvrd/devops-bench#14): it adds no Python and changes no behaviour, so it can land independently. `Dockerfile.openclaw` belongs to the openclaw image PR (staged as pradeepvrd/devops-bench#19) and is untouched here.

## Why the upstream installer is not piped to a shell

`curl -fsSL https://antigravity.google/cli/install.sh | bash` is the documented install path, and it is the wrong thing for an image:

- It resolves whatever the release manifest calls **latest**, which defeats pinning. A CLI bump would silently invalidate the A/B baseline a run was scored against.
- It installs into `$HOME` and rewrites shell profiles, both meaningless in an image whose `HOME` is container-owned and replaced per run.

What it does that is worth keeping is the sha512 check, so that is done inline against the pinned artifact: a tampered or truncated payload fails the build rather than quietly scoring a run.

Release paths are inconsistent between arches upstream (`linux-x64/cli_linux_x64.tar.gz` vs `linux-arm/cli_linux_arm64.tar.gz`), hence an explicit mapping rather than a substitution. The archive member is named `antigravity`; it is installed as `agy`, which is the name the harness spells.

`node:22-slim` rather than plain debian: agy is a Go binary and needs no runtime, but MCP server bindings are conventionally launched with `npx`, and a task declaring one would otherwise fail inside the boundary.

## Live validation

On the bastion, against the sandbox executor's real invocation — `--user <uid>:<gid>`, `HOME` on the workspace mount, and `GCE_METADATA_HOST` pointed at a dead port.

| stage | result |
| --- | --- |
| `DOCKER-USER` REJECT to `169.254.169.254` | present, rule 3 |
| metadata from inside a container | `http=000` |
| **control** — token removed from the mount | fails, `Authentication required` |
| **E2E** — OAuth token on the workspace mount | real completion |
| tool floor | `agy kubectl helm jq git rg node npx` all resolve |
| `npx` under the mapped uid | works |
| image size | 892 MB |

The control and the E2E differ only in whether the token is on the mount, so the pass exercises the credential path rather than an ambient one.

Two image requirements that were suspected and are **not** real, recorded so they are not re-litigated: the mapped uid needs no `/etc/passwd` entry (`getent passwd` returns `NONE` and agy does not care), and `node:22-slim` is missing no shared library agy needs.

## Not in this PR

- **The `_resolve_binary` fix.** `_resolve_binary()` returns the *host* path with no container-side counterpart to openclaw's `_CONTAINER_OC_BIN`, so a host path crosses into container argv. That belongs with the boundary that creates the problem and is held in the harness-wiring PR (staged as pradeepvrd/devops-bench#14).
- **The model-id bug.** The matrix sends agy `gemini-3.1-pro-preview` (`scripts/bastion/vm-setup.sh:209`); agy rejects the `-preview` suffix and requires a reasoning tier, so every agy run exits 1 at startup. Not sandbox-specific — an ambient run fails identically, and it is present on `main` today. Fixed in the companion PR #160.

Neither blocks this image; both are needed before an agy arm scores.

## Summary by CodeRabbit

* **New Features**
  * Added a container image for running the Antigravity CLI agent harness.
  * Includes preconfigured command-line tools such as kubectl, Helm, Git, jq, ripgrep, and curl.
  * Includes the Antigravity CLI and supports architecture-specific builds with verified tool downloads.
  * Uses pinned tool versions for consistent environments.
  * Configures `/workspace` as the working directory without embedding authentication or workspace data.

### DIFF
--- a/docker/Dockerfile.antigravity
+++ b/docker/Dockerfile.antigravity
@@ -1,0 +1,94 @@
+# Copyright 2026 The Kubernetes Authors.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+# Sandbox image for the Antigravity CLI agent harness (BENCH_SANDBOX_IMAGE).
+#
+# One image per harness: a CLI version bump here must not invalidate the A/B
+# baselines of the other agents under test, so the images are deliberately not
+# a shared multi-stage build with per-agent targets.
+#
+# Contents are the tools a task may legitimately need inside the boundary. The
+# agent's cluster credential, workspace, and fixtures are mounted at run time
+# by devops_bench/agents/sandbox.py — never baked in here. Nor is the OAuth
+# token the CLI authenticates with: the harness copies that into the per-run
+# workspace and points agy at it with --gemini_dir, so it arrives on the
+# workspace mount and is deleted when the run ends.
+
+# node, not plain debian: agy itself is a static-ish Go binary and needs no
+# runtime, but MCP server bindings are conventionally launched with npx, and a
+# task that declares one would otherwise fail inside the boundary.
+FROM node:22-slim
+
+# Versions are pinned so a rebuild reproduces the image an A/B arm was scored
+# with; bump them deliberately and re-baseline.
+ARG KUBECTL_VERSION=v1.31.0
+ARG HELM_VERSION=v3.16.3
+# AGY_BUILD is the build id that follows the version in the release path; the
+# two together address one immutable artifact.
+ARG AGY_VERSION=1.2.0
+ARG AGY_BUILD=5210873191596032
+ARG AGY_SHA512_AMD64=d190b25a04ed2a03b0587838476494ae59d0d324ee6373263cabe584352c254f9051f3689c79ac4add8817ff6eb34b11694acd77002c55e91bd1d28b6dfdae22
+ARG AGY_SHA512_ARM64=5ad72fba8c8e915c59c5505dc99116058352b8eba7f141d7ed220256788bada30df8b5af65c0d931ff10bbc661c0f5fe83931c58071fe3063a809bbeda7f59f9
+
+# Modern Git comes from backports for reftable support; ripgrep and jq are
+# what the agent's own search and JSON handling reach for.
+RUN echo "deb http://deb.debian.org/debian bookworm-backports main" > /etc/apt/sources.list.d/backports.list \
+    && apt-get update && apt-get install -y --no-install-recommends \
+        ca-certificates \
+        curl \
+        jq \
+        ripgrep \
+    && apt-get install -y -t bookworm-backports --no-install-recommends git \
+    && rm -rf /var/lib/apt/lists/*
+
+# kubectl and helm: resolve the arch from the build platform rather than
+# hardcoding amd64, so the image builds on an arm64 laptop as well as the
+# x86 bastion.
+RUN arch="$(dpkg --print-architecture)" \
+    && curl -fsSLO "https://dl.k8s.io/release/${KUBECTL_VERSION}/bin/linux/${arch}/kubectl" \
+    && install -o root -g root -m 0755 kubectl /usr/local/bin/kubectl \
+    && rm kubectl \
+    && curl -fsSL "https://get.helm.sh/helm-${HELM_VERSION}-linux-${arch}.tar.gz" \
+        | tar -xz -C /tmp \
+    && install -o root -g root -m 0755 "/tmp/linux-${arch}/helm" /usr/local/bin/helm \
+    && rm -rf "/tmp/linux-${arch}"
+
+# The upstream bootstrapper (https://antigravity.google/cli/install.sh) is
+# deliberately not piped to a shell here. It resolves whatever the release
+# manifest calls *latest*, which defeats the pinning above; it also installs
+# into $HOME and then rewrites shell profiles, both meaningless in an image
+# whose HOME is container-owned and replaced per run. What it does that we do
+# keep is the sha512 check, so a tampered or truncated payload fails the build
+# rather than silently scoring a run.
+#
+# Release paths are inconsistent between arches upstream (linux-x64 vs
+# linux-arm, cli_linux_x64 vs cli_linux_arm64), hence the explicit mapping
+# rather than a substitution. The archive member is named `antigravity`; it is
+# installed as `agy` because that is the name the harness invokes and, inside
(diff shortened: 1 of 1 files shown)


isadominguez314:
@coderabbitai review

docker/Dockerfile.antigravity
```diff
@@ -0,0 +1,94 @@
...
+# agent's cluster credential, workspace, and fixtures are mounted at run time
+# by devops_bench/agents/sandbox.py — never baked in here. Nor is the OAuth
+# token the CLI authenticates with: the harness copies that into the per-run
+# workspace and points agy at it with --gemini_dir, so it arrives on the
+# workspace mount and is deleted when the run ends.
+
+# node, not plain debian: agy itself is a static-ish Go binary and needs no
+# runtime, but MCP server bindings are conventionally launched with npx, and a
+# task that declares one would otherwise fail inside the boundary.
+FROM node:22-slim
```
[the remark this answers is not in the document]
isadominguez314: Took the part of this that strengthens a real guarantee and declined the rest. The base image is now pinned by its multi-arch manifest-list digest, and kubectl/helm gained pinned per-arch SHA-256 verification (see the sibling thread), so every layer that defines the arm under test — base, cluster tools, agent CLI — is now immutable under an unchanged Dockerfile.

Declining the Debian snapshot mirror for the apt layer. The comment this finding quotes was overpromising, and I've reworded it rather than chased byte-reproducibility: a scored A/B baseline is identified by the digest of the built image the arm actually ran (the harness pins and pre-pulls that image), not by the ability to rebuild identical bytes later — which docker builds don't deliver anyway. The apt packages involved are support tooling (git/jq/ripgrep), and having them track Debian security updates on deliberate rebuilds is the behavior we want; snapshot.debian.org would add a slow, occasionally unavailable moving part to every build without strengthening the guarantee that matters.

isadominguez314:
@coderabbitai review

docker/Dockerfile.antigravity
```diff
@@ -0,0 +1,114 @@
...
+
+# Modern Git comes from backports for reftable support; ripgrep and jq are
+# what the agent's own search and JSON handling reach for.
+RUN echo "deb http://deb.debian.org/debian bookworm-backports main" > /etc/apt/sources.list.d/backports.list \
+    && apt-get update && apt-get install -y --no-install-recommends \
+        ca-certificates \
+        curl \
+        jq \
+        ripgrep \
+    && apt-get install -y -t bookworm-backports --no-install-recommends git \
```
eugeneng04: bookworm-backports doesn't have a git package, so this installs the stock 2.39.5 from bookworm (checked with apt-cache policy git in the built image). 

2.39 has no reftable support, so the comment on L54 doesn't hold. I think we should drop the backports source on L56 and fix the comment, unless a task needs a newer git.

docker/Dockerfile.antigravity
```diff
@@ -0,0 +1,114 @@
...
+# an unchanged Dockerfile; bump the digest together with the tool pins below.
+FROM node:22-slim@sha256:83f487e0a63425e5b4d146fb5e5be574bcbe1b7b843d3ebafdd95eaf7767a7e5
+
+# Everything that defines the arm under test is pinned and checksummed: the
+# base image by digest above, the cluster tools and the agent CLI by version +
+# sha below. Bump them deliberately and re-baseline. The apt layer's support
+# packages (git/jq/ripgrep) deliberately track Debian security updates
+# instead — a scored baseline is identified by the digest of the built image
+# it actually ran, not by a byte-identical rebuild.
+ARG KUBECTL_VERSION=v1.31.0
```
eugeneng04: kubectl v1.31.0 is past end of support, and since this pin becomes the baseline runs are scored against, we might want to bump it before freezing. Same goes for helm v3.16.3 on L44.

maybe consult Pradeep or Simran on what version we should use

docker/Dockerfile.antigravity
```diff
@@ -0,0 +1,114 @@
...
+# the agy tarball below — a tampered or truncated payload fails the build
+# rather than silently scoring a run. The pins are per-arch because the
+# vendors publish per-arch sums.
+RUN arch="$(dpkg --print-architecture)" \
+    && case "${arch}" in \
+        amd64) kubectl_sha="${KUBECTL_SHA256_AMD64}"; helm_sha="${HELM_SHA256_AMD64}" ;; \
+        arm64) kubectl_sha="${KUBECTL_SHA256_ARM64}"; helm_sha="${HELM_SHA256_ARM64}" ;; \
+        *) echo "unsupported architecture for kubectl/helm: ${arch}" >&2; exit 1 ;; \
+    esac \
+    && curl -fsSLO "https://dl.k8s.io/release/${KUBECTL_VERSION}/bin/linux/${arch}/kubectl" \
```
eugeneng04: this downloads kubectl into `/` since there's no `WORKDIR` yet, while helm and agy go to `/tmp`. Maybe use `-o /tmp/kubectl` so all three are the same?

docker/Dockerfile.antigravity
```diff
@@ -0,0 +1,114 @@
...
+# a shared multi-stage build with per-agent targets.
+#
+# Contents are the tools a task may legitimately need inside the boundary. The
+# agent's cluster credential, workspace, and fixtures are mounted at run time
+# by devops_bench/agents/sandbox.py — never baked in here. Nor is the OAuth
+# token the CLI authenticates with: the harness copies that into the per-run
+# workspace and points agy at it with --gemini_dir, so it arrives on the
+# workspace mount and is deleted when the run ends.
+
+# node, not plain debian: agy itself is a static-ish Go binary and needs no
```
eugeneng04: Do we need all of these comments? A lot of it repeats the PR description. Also is agy static or not? "static-ish" doesn't tell us whether it needs any libs from the base.

eugeneng04 (changes requested):
we should probably update the kubectl and helm versions right? Unless theres a reason why we are using these specific versions.

Also I think there are a lot of overly verbose comments that make this really hard to read.

docker/Dockerfile.antigravity
isadominguez314: I'll message Pradeep and Simran. I've left them like this for now because kubectl only supports one minor version of skew against the API server. The kind-based tasks all pin kindest/node:v1.30.0, so today's v1.31 is within range for them, and the current stable (v1.37.1) would be six minors out. GKE clusters use the release-channel default, which is much newer. No single kubectl minor covers both right now, so the pin should follow whatever version we standardize the task clusters on.
For helm, the current v3 line is v3.22.0. v4.3.0 is also out, but it has breaking changes, so I'd stay on v3 unless a task needs v4.

PR #162, opened 2026-09-11
fix(bastion): pin openclaw to a version that accepts the harness config

isadominguez314:
One-line pin bump plus a known-issues row.

> Staged on the `pradeepvrd/devops-bench` integration fork as pradeepvrd/devops-bench#23 (merged there); the incident below was hit and verified in that environment.

## The break this prevents

#149 makes the harness write `memory.search.enabled` into every per-run `openclaw.json`, so runs cannot recall each other's transcripts. That key does not exist before openclaw **2026.8.1**. The bastion pin in `tf/modules/bastion/startup.sh` is **2026.6.10**, unchanged since it was introduced.

On `main` today the two have not collided yet — the config write arrives with #149. On the integration staging fork, where #149's change is already merged, the old pin broke **every** openclaw run: older `oc` rejects the *whole* config rather than ignoring an unknown key:

```
OpenClaw config is invalid ... memory: Invalid input
```

The agent then exits 1 having done nothing. The pin bump is forward-compatible (newer `oc` accepts today's configs), so this can land before #149; it must land no later. Bumping here rather than inside #149 keeps that PR reviewable as harness behaviour and this one as infra.

## Why it is easy to miss

The run does not report as an error. It reports as a success:

- `status: "success"`, `validated: false`, empty trajectory, 0 tools, ~15s
- every objective reads `not observed`
- correctness is withheld, so `_finalize_outcome_score` returns before emitting `OutcomeScore`

The record scores null and drops out of the leaderboard. Nothing in the summary says the agent never ran. I lost two matrix runs to this before finding the `oc` stderr.

## The fix

Bump the pin to `2026.8.2` and say in a comment that the two move together:

```
# 2026.8.x is a floor, not a preference: the harness writes `memory.search.enabled`
# into every per-run openclaw.json ... Bump the two together or not at all.
OPENCLAW_VERSION="2026.8.2"
```

## Operator action required — Terraform alone will not fix a running bastion

`startup.sh` only installs `oc` at VM creation, and it short-circuits on `/var/lib/bench-bastion-ready`. Existing bastions need:

```bash
sudo npm install -g openclaw@2026.8.2
```

Verified on a live bastion: the config that produced `memory: Invalid input` now reports `Config valid`, and a subsequent run reached the model call.

The known-issues row documents both the signature and the in-place fix, since anyone hitting this will search for the symptom rather than the cause.

## Summary by CodeRabbit

* **Bug Fixes**
  * Updated newly provisioned virtual machines to use OpenClaw 2026.8.2, preventing failures caused by unsupported per-run memory configuration.
  * Existing virtual machines may require a manual OpenClaw upgrade or VM rebuild to receive the fix.

* **Documentation**
  * Clarified that OpenClaw 2026.8.1 or later is required for the per-run memory configuration.
  * Documented the known failure mode and upgrade guidance.


isadominguez314:
@coderabbitai review

tf/modules/bastion/startup.sh
```diff
@@ -34,7 +34,14 @@ TOFU_VERSION="1.8.8"
...
 # new release rather than tracking @latest.
-OPENCLAW_VERSION="2026.6.10"
+#
+# 2026.8.x is a floor, not a preference: the harness writes `memory.search.enabled`
+# into every per-run openclaw.json to keep runs from recalling each other's
+# transcripts, and that key does not exist before 2026.8.1. On the pinned 2026.6.10
+# oc rejects the whole config -- `OpenClaw config is invalid ... memory: Invalid
+# input` -- and the agent exits 1 having done nothing, which scores as a null row
+# rather than an error. Bump the two together or not at all.
+OPENCLAW_VERSION="2026.8.2"
```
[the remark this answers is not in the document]
isadominguez314: Good catch on the tense, but the producer does exist — it's just not on `main` yet. The `memory.search.enabled` write lives in #149 (`_build_openclaw_config` sets `payload["memory"] = {"search": {"enabled": False}}`), which is already merged on the integration staging fork where this incident was hit; this PR's body describes that ordering ("the config write arrives with #149... can land before it; must land no later").

What *was* wrong is that the `startup.sh` comment and the known-issues row stated the write in the present tense, as if it were already in this tree — misleading for anyone reading `main` between the two merges. Fixed in 1dfb9e8: both now attribute the `memory` section to the incoming #149 change and note where the incident actually occurred. The pin itself is unchanged (2026.8.2, forward-compatible).

Not adding a `memory` producer here per the suggested alternative — that behavior change is exactly what #149 is, and it stays reviewable as harness behavior while this PR stays infra.

isadominguez314:
@coderabbitai review

isadominguez314:
@coderabbitai review

tf/modules/bastion/startup.sh
```diff
@@ -34,7 +34,18 @@ TOFU_VERSION="1.8.8"
 NODE_MAJOR="22"
 # Pin openclaw so VM rebuilds are reproducible; bump deliberately when adopting a
 # new release rather than tracking @latest.
-OPENCLAW_VERSION="2026.6.10"
+#
+# 2026.8.1 or newer is a floor, not a preference: an incoming harness change
```
eugeneng04: seems like a lot of comments for a version update

docs/appendix/known_issues.md
```diff
@@ -22,6 +22,7 @@ If an eval fails, find the symptom below and apply the action. Many failures are
...
+| oc exits 1 in ~15s having done nothing: `OpenClaw config is invalid ... Problem: - memory: Invalid input`. The record still reports `status: "success"`, but with `validated: false`, empty `trajectory`/`tools`, `cheating_report: no_data`, every objective `not observed`, and no `OutcomeScore` | The per-run `openclaw.json` gains a `memory` section (`memory.search.enabled: false`, so runs cannot recall each other's transcripts) with the harness change staged in PR #149 — already merged on the integration staging fork, where this incident occurred. That section only exists from openclaw **2026.8.1** (where the memory subsystem ships) on, and older oc rejects the whole config rather than ignoring the key. The bastion pin was left at `2026.6.10`, so that config and the VM disagreed | Fixed — the pin is now `2026.8.2`. **Terraform only installs oc at VM creation**, so an existing bastion does not pick this up: upgrade in place with `sudo npm install -g openclaw@2026.8.2`, or rebuild the VM. Confirm with `oc --version` | **Config / setup** | Yes |
```
eugeneng04: A few of these don't exist on main, so someone matching their record against this row won't find them. cheating_report isn't anywhere in the repo, objective statuses are only pass/fail/error, and validated is the task-vetted flag. I think we should keep just the oc error, status: "success", empty trajectory and missing OutcomeScore.

eugeneng04 (approved):
/lgtm

PR #163, opened 2026-09-11
fix(agents): default the CLI harnesses' Vertex location to global

isadominguez314:
> Staged on the `pradeepvrd/devops-bench` integration fork as pradeepvrd/devops-bench#24.

## The bug

`gemini_cli` and `antigravity` both defaulted an unconfigured Vertex run to `us-central1`:

```python
overlay["GOOGLE_CLOUD_LOCATION"] = (
    os.environ.get("GOOGLE_CLOUD_LOCATION")
    or os.environ.get("GCP_LOCATION")
    or "us-central1"
)
```

Every model id this benchmark routes at Vertex carries the `-preview` suffix, and those ids are published **only on the global endpoint** — a regional one answers `404 Publisher model ... not found`. `docs/appendix/known_issues.md` already says so:

> Use the **`global`** location (`GOOGLE_CLOUD_LOCATION=global` / `GCP_VERTEX_LOCATION=global`) and a `-preview` model id

So the default was the one value that cannot work. It also disagreed with `models/gemini.py`, `models/claude.py`, and the `claude_code` harness, which all default to `global`.

`scripts/bastion/_matrix_lib.sh:280` **already works around this at the shell layer**, exporting `GOOGLE_CLOUD_LOCATION=global` for Vertex matrix runs. That covers the matrix path only. A direct `run.py` invocation, a dev laptop, CI, and the sandboxed agent path all still got `us-central1`.

## The second half

Neither harness read `GCP_VERTEX_LOCATION` — the spelling the docs and the model adapters use. An operator who set the documented variable got `us-central1` anyway, silently.

## The fix

One shared `agents/shared/vertex_env.vertex_location()`, consumed by both harnesses, so the chain cannot drift apart again:

```
GOOGLE_CLOUD_LOCATION  ->  GCP_VERTEX_LOCATION  ->  (agy only: gcloud compute/region)  ->  "global"
```

### `GCP_LOCATION` is dropped, not demoted

I first demoted it below `GCP_VERTEX_LOCATION`. That wasn't enough. **Every** place this repo sets `GCP_LOCATION` sets a cluster **zone**:

| Source | Value |
| --- | --- |
| `scripts/bastion/vm-setup.sh` | `export GCP_LOCATION="us-central1-a"` — in the bastion profile |
| `scripts/bastion/_matrix_lib.sh` | `${GCP_LOCATION:-us-central1-a}` |
| `deployers/factory.py:30` | `_DEFAULT_LOCATION = "us-central1-a"` |
| `docs/components/infra.md` | documented as "Default region/zone" |

A zone is not a valid Vertex location, so reading it here could only ever route a run at an endpoint that does not exist. And because it is *always* set on the bastion, the corrected default was unreachable there. Measured on this branch:

```
GCP_LOCATION=us-central1-a, nothing else set
  demoted  -> us-central1-a      (invalid endpoint)
  dropped  -> global             (correct)
```

Operators pinning a Vertex region set `GCP_VERTEX_LOCATION`.

antigravity still **writes** `GCP_LOCATION` into its subprocess overlay — the `agy` binary's own GCP tooling has always been handed that spelling, so removing it changes the binary's inputs rather than fixing routing. That write only reaches the `agy` subprocess, never the deployers. Happy to drop it too if reviewers prefer.

### The gcloud fallback stays lazy

Passed as a callable, not a value, so a configured host never pays for the subprocess. Covered by a test.

## Deliberately not touched

`claude_code` already defaults to `global`, and its chain prefers the repo var over the native `CLOUD_ML_REGION` — the opposite order from these two. Reconciling that is a separate question from fixing a default that is wrong on its face, and that file is contended by other open PRs.

## Validation

- Unit suite green on this branch (1373 passed), ruff check and format clean. New `tests/unit/agents/shared/test_vertex_env.py` covers precedence, the zone-is-not-read rule, whitespace-only values, and fallback laziness. Two antigravity tests assert the wired behaviour end to end through `extra_env`.
- The **premise** was confirmed live on the bastion: `gemini-3.1-pro-preview` 404s at `us-central1`, returns 200 at `global`.
- **The default path has now run live** (2026-09-16, bastion): one ambient gemini_cli Vertex run on this branch's tip (`d481fd5`) with **both** `GOOGLE_CLOUD_LOCATION` and `GCP_VERTEX_LOCATION` unset via `env -u`, every API key unset (ADC only), invoked directly through `python3 -m devops_bench` — not the matrix scripts, which re-inject the location and had kept this path unreachable on the bastion until now. Run `run_20260915_222600_p4-default-loc-20260915`: status success, 24 trajectory steps, judged scores, on `gemini-3.1-pro-preview` — an id only the global endpoint serves, so the completion is itself the proof that `vertex_location()`'s in-code fallback supplied `global`. (The 404-at-`us-central1` premise measured above stands in as the negative control.)

🤖 Generated with [Claude Code](https://claude.com/claude-code)

## Summary by CodeRabbit

- **New Features**
  - Added consistent Vertex location resolution across Gemini CLI and antigravity integrations.
  - Supports Google Cloud and Vertex-specific location settings, with a default of `global`.
  - Gemini CLI now configures Vertex routing and project settings when using the Vertex backend.
  - Cluster-zone settings are no longer used for Vertex routing.

- **Documentation**
  - Documented the `GOOGLE_CLOUD_LOCATION` environment variable and its precedence.

- **Bug Fixes**
  - Prevented non-Vertex runs from unintentionally inheriting settings that reroute traffic to Vertex.

### DIFF
--- a/devops_bench/agents/cli/antigravity/agent.py
+++ b/devops_bench/agents/cli/antigravity/agent.py
@@ -234,6 +234,10 @@ class AgyCliAgent(base.AgentHarness):
             # chain is empty.
             location = vertex_location(fallback=_get_gcloud_location)
             env_overlay["GOOGLE_CLOUD_LOCATION"] = location
+            # Written but deliberately not *read* back (see vertex_env): agy's
+            # own GCP tooling has always been handed this spelling, and dropping
+            # it is a behavior change for the binary, not a routing fix. It only
+            # ever reaches the agy subprocess, never the deployers.
             env_overlay["GCP_LOCATION"] = location
 
             # Explicit gemini_dir keeps agy on the workspace settings, not real HOME.
--- a/devops_bench/agents/cli/antigravity/agent.py
+++ b/devops_bench/agents/cli/antigravity/agent.py
@@ -29,6 +29,7 @@ from devops_bench.agents import config as agents_config
 from devops_bench.agents import result as agents_result
 from devops_bench.agents.cli.antigravity import parsing
 from devops_bench.agents.shared import cli_capabilities
+from devops_bench.agents.shared.vertex_env import vertex_location
 from devops_bench.core import subprocess as devops_subprocess
 
 if TYPE_CHECKING:
@@ -228,15 +229,12 @@ class AgyCliAgent(base.AgentHarness):
                 env_overlay["GOOGLE_CLOUD_PROJECT"] = project
                 env_overlay["GCP_PROJECT"] = project
 
-            location = (
-                os.environ.get("GOOGLE_CLOUD_LOCATION")
-                or os.environ.get("GCP_LOCATION")
-                or _get_gcloud_location()
-                or "us-central1"
-            )
-            if location:
-                env_overlay["GOOGLE_CLOUD_LOCATION"] = location
-                env_overlay["GCP_LOCATION"] = location
+            # Shared with the Gemini CLI harness so the two cannot drift; the
+            # gcloud lookup stays a lazy fallback, consulted only when the env
+            # chain is empty.
+            location = vertex_location(fallback=_get_gcloud_location)
+            env_overlay["GOOGLE_CLOUD_LOCATION"] = location
+            env_overlay["GCP_LOCATION"] = location
 
             # Explicit gemini_dir keeps agy on the workspace settings, not real HOME.
             argv = [
--- a/devops_bench/agents/cli/gemini_cli/agent.py
+++ b/devops_bench/agents/cli/gemini_cli/agent.py
@@ -47,6 +47,7 @@ from devops_bench.agents.shared.cli_capabilities import (
     build_mcp_servers,
     materialize_skills,
 )
+from devops_bench.agents.shared.vertex_env import vertex_location
 from devops_bench.core import SubprocessError, get_logger
 from devops_bench.core.model_providers import resolve_provider
 from devops_bench.core.subprocess import run
@@ -151,7 +152,13 @@ def _build_env(config: AgentConfig) -> dict[str, str]:
     ``config.model`` onto ``GEMINI_MODEL``. OTLP telemetry exporters are disabled
     so they don't hang on broken endpoints. The model is never hardcoded; it
     flows from ``config.model``. A keyless backend (e.g. Vertex/ADC) writes no
-    key.
+    key. A Vertex backend additionally writes the google-genai routing vars
+    (``GOOGLE_GENAI_USE_VERTEXAI`` plus project/location), since the SDK
+    otherwise talks to the Gemini API regardless of the configured provider. The
+    location comes from the shared
+    :func:`~devops_bench.agents.shared.vertex_env.vertex_location` chain, which
+    defaults to ``global`` — the only endpoint the ``-preview`` model ids are
+    published on.
 
     Args:
         config: Resolved :class:`AgentConfig` for this run.
@@ -173,6 +180,16 @@ def _build_env(config: AgentConfig) -> dict[str, str]:
         "OTEL_LOGS_EXPORTER": "none",
         "OTEL_SDK_DISABLED": "true",
     }
+    if spec.backend == "vertex":
+        # The google-genai SDK the CLI embeds reads these three; without the
+        # switch it defaults to the Gemini API and ignores the Vertex routing.
+        # Project/location env spellings follow the antigravity harness so an
+        # operator configures both agents the same way.
+        overlay["GOOGLE_GENAI_USE_VERTEXAI"] = "true"
+        project = os.environ.get("GOOGLE_CLOUD_PROJECT") or os.environ.get("GCP_PROJECT")
+        if project:
+            overlay["GOOGLE_CLOUD_PROJECT"] = project
+        overlay["GOOGLE_CLOUD_LOCATION"] = vertex_location()
     if config.api_key:
         for var in spec.api_key_envs:
             overlay[var] = config.api_key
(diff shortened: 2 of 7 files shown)


isadominguez314:
@coderabbitai review

devops_bench/agents/shared/vertex_env.py
```diff
@@ -0,0 +1,88 @@
...
+        default: Value returned when neither the env chain nor ``fallback``
+            yields anything.
+
+    Returns:
+        The first non-empty value from :data:`VERTEX_LOCATION_ENVS`, else
+        ``fallback()``, else ``default``. Values are stripped, so a variable set
+        to whitespace is treated as unset rather than routing traffic at ``" "``.
+    """
+    for name in VERTEX_LOCATION_ENVS:
+        value = (os.environ.get(name) or "").strip()
```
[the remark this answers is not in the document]
isadominguez314: The guideline's own carve-out covers this case: review instruction 2 says "Do not flag the model-provider axis (gemini, claude, ollama, Vertex AI): a model adapter naming its own backend or reading SDK-required credentials/project ID is correct." GOOGLE_CLOUD_LOCATION / GCP_VERTEX_LOCATION are Vertex AI model-endpoint routing variables — the model-provider axis — not cloud-infra resolution. The suggested destination doesn't fit either: devops_bench/providers/ is the infrastructure-provider layer (cluster provisioning), and moving model routing there would conflate exactly the two axes the guidelines separate. agents/shared/ holds this resolver because three vendor harnesses (gemini, antigravity, openclaw) share one location chain — duplicating it per vendor subtree trades a layering nit for three copies of routing logic that must never drift. If maintainers still want it out of the generic agents layer, the coherent home would be the model-provider seam in core/model_providers.py, not providers/ — happy to do that move in a follow-up if a human reviewer prefers it. Leaving as-is for now.

isadominguez314:
@coderabbitai review

devops_bench/agents/cli/gemini_cli/agent.py
```diff
@@ -173,6 +180,25 @@ def _build_env(config: AgentConfig) -> dict[str, str]:
         "OTEL_LOGS_EXPORTER": "none",
         "OTEL_SDK_DISABLED": "true",
     }
+    if spec.backend == "vertex":
+        # The google-genai SDK the CLI embeds reads these three; without the
+        # switch it defaults to the Gemini API and ignores the Vertex routing.
+        # Project/location env spellings follow the antigravity harness so an
+        # operator configures both agents the same way.
+        overlay["GOOGLE_GENAI_USE_VERTEXAI"] = "true"
+        project = os.environ.get("GOOGLE_CLOUD_PROJECT") or os.environ.get("GCP_PROJECT")
```
eugeneng04: The repo-wide project var is GCP_PROJECT_ID (claude_code, models/gemini.py, providers/gcp.py). I think we should add GCP_PROJECT_ID here, same as GCP_VERTEX_LOCATION was added for location

tests/unit/agents/test_agents_cli_gemini.py
```diff
@@ -194,6 +194,80 @@ def test_build_env_keyless_writes_no_key_var() -> None:
...
+    monkeypatch.setenv("GCP_VERTEX_LOCATION", "europe-west1")
+    env = _build_env(AgentConfig(model="gemini-2.5-pro", provider="google-vertex"))
+    assert env["GOOGLE_CLOUD_PROJECT"] == "proj-google"
+    assert env["GOOGLE_CLOUD_LOCATION"] == "asia-northeast1"
+
+
+def test_build_env_vertex_defaults_the_location_to_global(
+    monkeypatch: pytest.MonkeyPatch,
+) -> None:
+    # No project anywhere: write nothing rather than an empty string, and let
```
eugeneng04: The comment says no project lets "the SDK's own ADC-derived default apply", but gemini-cli validates that GOOGLE_CLOUD_PROJECT is set when GOOGLE_GENAI_USE_VERTEXAI=true and errors otherwise. So this test is asserting a config that can't run. Either fix the comment or have _build_env fail early with a clear error when there's no project

devops_bench/agents/cli/antigravity/agent.py
```diff
@@ -228,15 +229,16 @@ def _execute(
...
-                or _get_gcloud_location()
-                or "us-central1"
-            )
-            if location:
-                env_overlay["GOOGLE_CLOUD_LOCATION"] = location
-                env_overlay["GCP_LOCATION"] = location
+            # Shared with the Gemini CLI harness so the two cannot drift; the
+            # gcloud lookup stays a lazy fallback, consulted only when the env
+            # chain is empty.
+            location = vertex_location(fallback=_get_gcloud_location)
```
eugeneng04: `_get_gcloud_location` reads compute/region, so on a host with a gcloud region set this returns something like us-central1 and the -preview ids 404 again, same as before. It's the same problem as GCP_LOCATION, a Compute setting and not a Vertex one. Should we drop the fallback so it lands on global?

devops_bench/agents/cli/antigravity/agent.py
```diff
@@ -228,15 +229,16 @@ def _execute(
...
+            env_overlay["GOOGLE_CLOUD_LOCATION"] = location
+            # Written but deliberately not *read* back (see vertex_env): agy's
```
eugeneng04: i dont think we need this comment do we?

devops_bench/agents/cli/gemini_cli/agent.py
```diff
@@ -173,6 +180,25 @@ def _build_env(config: AgentConfig) -> dict[str, str]:
...
+        if project:
+            overlay["GOOGLE_CLOUD_PROJECT"] = project
+        overlay["GOOGLE_CLOUD_LOCATION"] = vertex_location()
+    else:
+        # The overlay rides on top of the inherited environment, so *omitting*
```
eugeneng04: I don't think we need this comment

eugeneng04 (approved):
/lgtm but address comments

PR #182, opened 2026-09-15
feat(sandbox): give the agent a scoped credential and make GKE and vcluster work

isadominguez314:
**Stacked on #181** (`up/sandbox-01-core-seam`). Base is `main` because a cross-fork PR cannot target a fork branch, so this diff includes #181's commits. **Incremental diff (this PR only): https://github.com/isadominguez314/devops-bench/compare/up/sandbox-01-core-seam...up/sandbox-02-credentials**

**⚠️ HOLD MERGE until #181 merges** — review can proceed in parallel; I'll rebase to a clean diff the moment #181 lands.

Makes the first **scoreable** sandboxed run possible. 1583 unit tests pass, ruff clean.

## Why this is needed

#181 built the boundary but could only reach **kind**, and it mounted the operator's **admin client certificate** as the agent's cluster credential. Two consequences made it unscoreable:

- `build_network_plan` raised `SandboxError` for any non-kind context, and scored runs must use remote-node clusters — co-located kind is dev-only. That is observed incident 1: a privileged pod on a kind node read the bench checkout off the host disk.
- GKE could not work in-container at all, because its kubeconfig authenticates through `gke-gcloud-auth-plugin`, which needs a `gcloud` binary and ADC the container deliberately lacks.

## What lands

**A scoped credential instead of the admin cert.** New `devops_bench/k8s/agent_credentials.py` creates a `bench-agent` ServiceAccount in `bench-system`, binds built-in `edit` cluster-wide plus a small cluster-scoped supplement, and mints a short-lived token (run timeout + slack, clamped to [10m, 2h]). The rendered kubeconfig carries that token and no `exec:` block. The admin path survives only behind `BENCH_SANDBOX_ALLOW_ADMIN_CREDS=1`; without it, a failure to mint raises rather than falling back.

**Per-provider network plans.** New `Provider.sandbox_network_plan` hook — default is a plain bridge, so a new provider gets a working sandbox for free and only kind carries special knowledge. `build_network_plan(provider, cluster_info)` applies the hook, then a generic loopback normalisation rewrites `127.0.0.1`/`localhost`/`::1` to `host.docker.internal`. GKE and vcluster stop raising.

**Pod security.** PSA `baseline` labels on provision-time namespaces, plus a `ValidatingAdmissionPolicy` backstop covering namespaces the agent creates *later* — which a label cannot, and which `deploy-hello-app` literally asks the agent to do. Two further policies guard the exemptions that policy needs, and a fourth covers pods that predate it (both below). New `agent_pod_security:` task key (`baseline` default, `privileged` opts out); an unrecognised value is a load-time error, not a silent fall-back.

**Kubernetes 1.30 is a floor.** `admissionregistration.k8s.io/v1` only reached GA there, and both `node_image` defaults still pointed at `kindest/node:v1.29.2` — so *every sandboxed kind task* failed on a module-built cluster, which validation on GKE and vcluster could not see. Raised to the digest-pinned `v1.30.0` five `tf/prebuilt` stacks already use, and the version is now a checked precondition rather than only a default: left to the apply it surfaces as `no matches for kind`, which reads like a typo in our own manifest. Deliberately a `SandboxError` and not routed through `BENCH_SANDBOX_ALLOW_ADMIN_CREDS` — that hatch is for an operator who cannot write cluster-scoped objects, and no credential makes a 1.29 apiserver serve a v1 policy. Thanks @pradeepvrd.

**The metadata endpoint is blocked** for containers on the bastion (`DOCKER-USER -d 169.254.169.254 -j REJECT`, DNS excepted — on a GCP VM the metadata address *is* the resolver, which kind conceals). That is observed incident 2. Minting a scoped model token is pointless while the container can curl the VM service account's unscoped one.

**Refusal to run unpinned.** If the plan carries no context pin — `--no-infra`, the no-op deployer — provisioning would write a cluster-wide admission policy and ClusterRoleBindings to whatever the ambient current-context names. It refuses and says which cluster it declined to touch.

**One ported prerequisite.** `_run_kubectl` gains an explicit `--context` pin (final commit, co-authored with @geojaz — ported from the fork's integration branch, where it rode the http_probe verification rework). The credential provisioning applies cluster-scoped RBAC and reads it back; doing either against whichever cluster the ambient current-context names is the difference between scoping the agent down and handing it someone else's credential. This overlaps the verification work migrating upstream separately (#146–#148 and the staging PRs behind them); whichever lands first, the other reconciles trivially — the function bodies are identical.

## Interim state, please read

**A sandboxed Vertex run cannot authenticate off this PR alone.** The metadata block plus the matrix runner unsetting API keys means Vertex falls back to ADC through exactly the blocked endpoint. It **fails loud** rather than degrading to an unsandboxed run, which is correct under that invariant — but it means sandboxing is usable with `AGENT_PROVIDER=google` + `AGENT_API_KEY` and **unavailable for the matrix default** until #183, stacked on this PR. **Ambient runs are unaffected by the block** — they are host processes, so `DOCKER-USER` is never consulted for them.

## Operator prerequisite on GKE

The runner identity needs **`roles/container.admin`**. GKE gates admission-policy resources behind `container.thirdPartyObjects.*`, which `roles/container.developer` lacks — without it the pod-security backstop silently is not installed. Measured, not assumed: this blocked validation until the grant was made.

## Known residuals

Documented in `docs/components/agents.md` rather than left implicit:

- `edit` bound cluster-wide reaches system namespaces: `impersonate` on serviceaccounts and `create` on serviceaccounts/token, so the agent can mint a token for a ServiceAccount bound to `cluster-admin`. A path to cluster-admin, and RBAC has no deny rule, so nothing in the supplement can subtract it. The `pods/exec` route alongside it is now denied by the exempt-namespace guard, but that is a patch over one exit rather than a fix for the scope.
- `edit` carries read and write on `secrets` in every namespace. Inherited from the built-in role, not narrowed anywhere.
- The supplement grants `update`/`patch` on namespaces, so PSA labels can be stripped. The admission policies are unaffected, so this removes the non-load-bearing half.
- **Nothing is torn down.** `bench-system`, the bindings, the admission policies and the PSA labels outlive the run. Fine on a disposable cluster; on a reused one the surviving policy applies to the operator too and can deny the next run's fixtures (`opa-remediation` ships two privileged pods). Teardown is the lifecycle PR that follows this stack — flagging it rather than letting someone rediscover it as a flaky fixture.
- `edit` covers Kubernetes' own API groups, not installed CRDs: Kyverno aggregates policy writes to `admin` only, so **`opa-remediation` cannot be fully passed under the default scope** (two of its objectives flip policies from `Audit` to `Enforce`) and its sandboxed score is not comparable to its ambient one. Nothing in a trajectory reveals this — no `forbidden` appears, the objective reads as an agent miss. The fix is a per-task RBAC supplement, its own PR, since it lets the artifact under test widen its own sandbox and deserves review on that basis.

## Validation

Run live on a GCP bastion, docker only.

- **A scored run on plain GKE** — the case #181 could not run at all. Every deterministic remediation objective passed, all three policy reports clear, all four safeguards held, **zero** `forbidden`/`denied` across 41 trajectory steps.
- **A scored run on vcluster**, same shape.
- **The scoped token is self-proving on GKE**: its kubeconfig has no `exec:` plugin and the container has no `gcloud`, so an accidentally-mounted ambient config would have failed every kubectl. They all worked.
- **Metadata block proven three ways**: container blocked (exit 7), host still working (exit 0 + token), rule present in the chain.
- **18/18 boundary probes green on GKE and 15/15 on kind**, controls first so a dead token cannot make every denial "pass". Both observed escapes denied. The probes found and closed two real holes during validation — an admitted `kubectl run --privileged -n kube-system` (closed by the exempt-namespace guard, denying the agent's own username any pod-producing write or exec in exempt namespaces) and pre-provisioning pods that admission never saw (closed by the nonconformant-pod shell guard, since a `pods/exec` CONNECT carries no object to test). Each guard is proven discriminating, not just failing closed: the conformant twin beside each denied probe succeeds.
- **vcluster host-isolation measured**: the agent token authenticates against the vcluster apiserver (200) and is rejected by the host apiserver (401), control first.
- **Flag-off parity, measured**: base and branch each ran `opa-remediation` ambient; all 15 deterministic verification entries identical, judge-scored items moved in both directions (n=1 per arm — the deterministic match is the load-bearing evidence).

The probe tooling itself is dev tooling and deliberately not shipped here; the stack's final PR promotes the probes into an e2e regression suite.

## Summary by CodeRabbit

* **New Features**
  * Added opt-in Docker sandboxing with isolated workspaces, credentials, fixtures, environment controls, and provider-specific network access.
  * Added scoped Kubernetes identities, pod-security controls, task-level security settings, and context-aware operations.
  * Added Vertex AI environment routing and Gemini CLI sandbox execution.
  * Added sandbox-aware cheat-detection rules for mounted fixtures.

* **Security**
  * Sandbox failures now stop affected runs instead of falling back to host execution.
  * Blocked container access to cloud metadata credentials while preserving DNS functionality.

* **Documentation**
  * Documented sandbox configuration, security boundaries, infrastructure requirements, and task settings.


isadominguez314:
**Acceptance smoke: validated live on this branch's tree.** Run 2026-09-16 on the GCP bastion (Linux, docker 29.1.3), single worktree at `up/sandbox-025-vertex-metadata` tip `049ed20` — which contains this PR's `1e4a56b`, so the credential path below is exactly this PR's code. Details of the shared setup in the #181 comment.

What exercised this PR specifically:

1. **The scoped credential, through the real path**: the 18/18 probe run provisions via `provision_agent_credentials` — `bench-agent` SA minted, token-bearing kubeconfig rendered, and the four controls proving the credential *works* before any deny is trusted. All five admission-policy guards fired where they should (`bench-agent-pod-security` denied the privileged and hostPath pods with its own message; the exempt-namespace guard denied by name; the nonconformant-pod guard refused exec into the pre-provisioning privileged pod while admitting exec into its conformant twin). Teardown left no residue.
2. **A scored sandboxed run** (`run_20260916_190617_ups3a-hello`, deploy-hello-app, key-based provider): success with judged scores — the end-to-end shape this PR makes possible. Kind co-located (dev-only per the invariants); the remote-node GKE/vcluster scored runs remain covered by the fork-side validation this tree was re-cut from, byte-identical in the security-bearing files.
3. The #183 emulator runs (positive + negative) authenticated to the cluster with this PR's scoped token while the model credential came from the emulator — the two credential paths compose.

isadominguez314:
@coderabbitai review

isadominguez314:
@coderabbitai review

isadominguez314:
@coderabbitai review

PR #189, opened 2026-09-16
feat(agents): run a task's turns as one conversation

jessie1111101:
## What

A task can only ever say one thing to the agent. This lets it hold a conversation.

`turns:` is a new optional task field — follow-up texts sent after `prompt`, in order, **within the same agent session**:

```yaml
prompt: "Diagnose why {{TARGET_DEPLOYMENT_NAME}} is failing in {{NAMESPACE}}."
turns:
  - "What would you change to stop it recurring?"
```

Placeholders resolve in a turn exactly as in the prompt. The whole exchange folds into one `AgentResult` — one trajectory, one token total, one latency — so scoring, reporting, and the leaderboard see one run, as they do today.

`AgentHarness.run_turns(prompts)` is the entry point. The `adk` harness overrides it to drive every turn through a single ADK `Runner` session.

## Why

Two reasons, and the second is the load-bearing one.

An agent whose behaviour *is* a conversation — clarify, then act, then follow up — can't be benchmarked by a single prompt at all.

And for a **remote** agent, one session is the only way to reach it coherently. An A2A service keys its conversation on `ContextId`, and `RemoteA2aAgent` derives that `ContextId` from the ADK session. Sending turn 2 as a second run creates a second session, hence a second context, and the remote answers having forgotten turn 1. Keeping one session is the entire mechanism; there is no A2A-level code here because none is needed.

## Compatibility

Nothing about an existing task moves. `turns` defaults to empty and every task in the repo omits it, so each one still takes `run()` on the unchanged single-turn path. `execute_agent` only reaches for `run_turns` when a task actually asks for more than one turn.

<details>
<summary><b>Why the base <code>_execute_turns</code> is not a loop over <code>_execute</code></b></summary>

The obvious default — call `_execute` once per turn and concatenate — is worse than no support at all. A harness that holds no session would restart the conversation on every turn, so the agent answers turn *n* having forgotten turns 1..n-1. The result is a transcript that reads like a conversation, scores like a conversation, and means nothing; the failure is invisible in the artifacts.

So the base errors instead: *"X is single-turn, but the task asked for N turns; it cannot hold a session across them."* A harness opts in by overriding `_execute_turns`. Today only `adk` does.

</details>

<details>
<summary><b>Reviewer notes</b></summary>

- **`timeout_sec` is the budget for the whole conversation, not per turn.** It bounds the agent's total wall-clock cost; a per-turn budget would let an N-turn task quietly cost N times what the same configured budget buys a single-turn one.
- **`run()` and `run_turns()` share one guard.** The latency stamp and the crash-to-errored-result safety net moved into `_guarded`, rather than being duplicated — behaviour is unchanged for `run()`, and the only visible difference is a log message that now says "agent execution" instead of "agent `_execute`".
- **`turns` is stripped and strictly validated like `prompt`.** A non-list, or a list with a non-string element, raises at load time rather than reaching the agent.

</details>

## Testing

`uv run pytest -q` — 1465 passed.

The multi-turn ADK test drives a **real** ADK agent whose stub model reports, on each turn, how many user turns it was handed: the run records `[1, 2]`. A fresh session per turn would record `[1, 1]`, so the assertion is evidence of session reuse rather than a mock echoing back the session id it was given.

Also covered: the single-turn default erroring on a multi-turn task, a one-element sequence behaving exactly as `run()`, an empty sequence, `run_turns` keeping the latency stamp and the safety net, `execute_agent` staying on the single-turn path when a task has no turns, placeholder substitution reaching every turn, and the schema's parse/strip/reject paths.

`uv run ruff check devops_bench tests` and `ruff format --check` clean.


jessie1111101:
@coderabbitai rate limit

PR #196, opened 2026-09-18
Add display metadata to tasks and verification entries

pradeepvrd:
## Why

A result viewer cannot tell what a task asks for or what a failing check means. The only task-level display field is the `name` slug, and the meaning of a verification entry such as `web-cpu-limit` lives in YAML comments that never reach the result record. The runtime `reason` is machine text ("path resolved to 0 values"), not an explanation.

This is the task-schema half. A follow-up stacked on this branch carries the metadata onto the result record and the leaderboard row.

## What

**Task** gains display-only fields:

| Field | Purpose |
| --- | --- |
| `title` | Display name; free to change, `name` stays the identity |
| `summary` | A few plain sentences: starting state, what to do, what done looks like |
| `category` | One bucket for filtering (`deploy`, `remediate`, `scale`, `secure`, `incident`, `migrate`, `generate`) |
| `tags` | Secondary facets |
| `check_groups` | `{slug: {title, description}}`, the requirement buckets entries opt into via `group` |

**VerificationEntry** gains `title`, `description`, `group`, `failure_hint`. None of it affects scoring or matching; `name` remains the chaos `verify:` key.

**Rules**, enforced on `Task`:

- Display text is never placeholder-substituted, so `{{...}}` in it is rejected.
- A `group` must be declared under `check_groups`.
- Once `validated: true`, the task-level fields and a `title` and `description` on every entry are required. The leaderboard renders validated tasks only, so that is where missing metadata should fail loudly. Unvalidated tasks stay loadable without any of it, so tasks in flight elsewhere are unaffected until promoted.

Both in-repo tasks are migrated (they were also the source of the implicit groups: the "weight categories" comment blocks are now data). `add-a-task.md` and the `task-review` checklist describe the fields.

## Rollout note

`VerificationEntry` is `extra="forbid"`, so a task carrying these keys on a harness that predates them lands in `verification_parse_errors`, which fail closed into the correctness score. Land this before any task that uses the fields.

## Deferred on purpose

- A structured judge rubric replacing free-text `expected_output`. Same need, but it touches the checklist metric parser, so it is its own change.
- `difficulty`.

## Testing

- `uv run pytest tests/unit`: 1455 passed, 11 skipped.
- Both in-repo tasks load through `Task.from_dict` and `parse_entries` with zero errors, and pass the `validated: true` rule.
- `ruff check` and `ruff format --check` clean.

## Summary by CodeRabbit

- **New Features**
  - Tasks now support viewer-facing titles, summaries, categories, tags, grouped checks, verification descriptions, and failure hints.
  - Added display metadata for the OPA remediation and GCP deployment tasks.

- **Bug Fixes**
  - Validated tasks now enforce required display metadata, approved categories, declared check groups, and placeholder-free text.
  - Display text is consistently trimmed, and blank group titles are rejected.

- **Documentation**
  - Expanded task-authoring guidance with metadata requirements, grouping instructions, valid categories, and examples.


tasks/common/opa-remediation/task.yaml
```diff
@@ -1,5 +1,26 @@
 task_id: 20
 name: "opa-remediation"
+title: "Remediate policy violations on a GitOps-managed cluster"
+summary: >
+  A kind cluster runs two Kyverno policies in Audit mode and four Deployments
+  violate them: two run privileged and two have no resource limits. Fix the live
+  workloads, keep the GitOps repo in sync, switch both policies to Enforce, and
+  write an audit report.
```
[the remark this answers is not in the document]
pradeepvrd: Keeping the clause. `summary` describes what the task asks the agent to do, and the prompt explicitly requires keeping the GitOps repo in sync; a viewer reading the summary should see the full ask. What is scored is expressed by the verification entries and their `check_groups`, which is where the missing `git_repo_sync` verifier is visible (and commented on inline). Dropping the clause from the summary would misdescribe the task rather than fix the coverage gap.

tests/unit/tasks/test_tasks_schema.py
```diff
@@ -251,3 +256,116 @@ def test_safety_checklists_empty_block_coalesces_to_empty_list():
...
+def test_validated_task_with_full_metadata_loads():
+    assert Task.from_dict(_validated_raw()).validated is True
+
+
+def test_validated_task_requires_task_level_fields():
+    with pytest.raises(ValidationError, match="requires summary, category"):
+        Task.from_dict(_validated_raw(summary="", category=""))
+
+
+def test_validated_task_requires_entry_title_and_description():
```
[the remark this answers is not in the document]
pradeepvrd: Leaving these unannotated on purpose: every existing test and helper in this file is unannotated, and mixing styles within one module is worse than matching it. Happy to annotate the whole file in a separate cleanup if maintainers want that convention here.

devops_bench/tasks/schema.py
```diff
@@ -174,6 +231,72 @@ def _coalesce_empty(cls, data: Any) -> Any:
...
+                raise ValueError(f"verification entry {label!r}: group must be a string")
+            if group not in self.check_groups:
+                raise ValueError(
+                    f"verification entry {label!r} names group {group!r}, "
+                    f"which is not declared under check_groups"
+                )
+
+        if not self.validated:
+            return self
+        missing = [f for f in ("title", "summary", "category") if not getattr(self, f).strip()]
```
eugeneng04, line 298: a validated task still loads with `check_groups: {g: {title: "   "}}` since this rule never looks at group titles, and group and entry text skips the `_text` strip the task fields get, so `"  "` survives verbatim. we might want to require group titles here and strip display text the same way everywhere.

tests/unit/tasks/test_repo_tasks.py
```diff
@@ -0,0 +1,46 @@
...
+
+from pathlib import Path
+
+import pytest
+
+from devops_bench.tasks.loader import _load_yaml_task, load_from_tasks_dir
+from devops_bench.verification.spec import parse_entries
+
+_TASKS_DIR = Path(__file__).resolve().parents[3] / "tasks"
+_TASK_FILES = sorted(_TASKS_DIR.glob("*/*/task.yaml"))
```
eugeneng04: `load_from_tasks_dir` discovers tasks recursively at any directory depth, but `glob("*/*/task.yaml")` is hardcoded to depth 2. Should we use `_TASKS_DIR.rglob("task.yaml")` so deeper nested tasks aren't skipped in this test?

tests/unit/tasks/test_repo_tasks.py
```diff
@@ -0,0 +1,46 @@
...
+        task_file, name_default=task_file.parent.name, folder=task_file.parent.name
+    )
+    assert task is not None
+    _, errors = parse_entries(task.verification_spec)
+    assert errors == []
+
+
+def test_directory_loader_drops_no_shipped_task() -> None:
+    loaded = {task.folder for task in load_from_tasks_dir(str(_TASKS_DIR))}
+    assert loaded == {p.parent.name for p in _TASK_FILES}
```
eugeneng04: both sides collapse to leaf directory names, so if `tasks/gcp/foo` and `tasks/common/foo` both exist and the loader drops one, the sets are still equal and this passes. Maybe compare sorted lists so a dropped duplicate shows up.

tasks/common/opa-remediation/task.yaml
```diff
@@ -1,5 +1,26 @@
...
+category: remediate
+tags: [kubernetes, kyverno, policy, gitops, security]
+# Display groups for the verification entries below. Each group's entry weights
+# sum to 1.0 so no group dominates the objective total; the group slug on each
```
eugeneng04, line 12: the comment says each group's weights sum to 1.0, but `cluster-compliant` is 6 x 0.1667 = 1.0002 and `policy-reports-clear` is 3 x 0.3333 = 0.9999, so `cluster-compliant` counts slightly more than stated. `deploy-hello-app` bumps its last entry to 0.3334, maybe do the same here.

tasks/gcp/deploy-hello-app/task.yaml
```diff
@@ -285,6 +357,9 @@ verification_spec:
   # needs the same pre-agent snapshot channel that `unchanged_outside` is waiting
   # on: verification currently runs as a single post-agent pass.
   - name: not-dumped-in-default
+    title: "Nothing deployed to the default namespace"
```
eugeneng04: Maybe align the title with the description "No deployment in default namespace" as a StatefulSet or bare Pod dumped in default namespace will pass this check.

tasks/gcp/deploy-hello-app/task.yaml
```diff
@@ -304,6 +379,8 @@ verification_spec:
   # innocent run destroys that run's score. Tighten this once the pre-agent
   # snapshot channel lands and the delta can be measured instead of approximated.
   - name: blast-radius
+    title: "Nothing deployed to kube-system"
```
eugeneng04: same as above, the check only checks for ones named "hello-app"

tasks/common/opa-remediation/task.yaml
```diff
@@ -217,6 +270,10 @@ verification_spec:
   # one `policy-reports-clear` all-node over these three checks; split so
   # each namespace's report state scores independently.
   - name: reports-clear-team-alpha
+    title: "team-alpha policy report is clear"
+    description: "The PolicyReport in namespace team-alpha lists no failing results."
```
eugeneng04, line 275: per the comment at the top of the spec in line 73, a namespace with no PolicyReport at all fails this on the emptiness guard, but the description and failure_hint only cover the failing-results case, so that failure reads as "a violation was not remediated". maybe add the missing-report case to the failure_hint.

eugeneng04 (approved):

devops_bench/tasks/schema.py
pradeepvrd: Fixed in a07cf3c: CheckGroup now strips title and description like task fields and rejects a blank title at any stage, and VerificationEntry strips its display fields on parse, so whitespace can neither satisfy the validated rule nor reach a record.

tests/unit/tasks/test_repo_tasks.py
pradeepvrd: Fixed in a07cf3c: switched to rglob so discovery matches the loader at any depth.

tests/unit/tasks/test_repo_tasks.py
pradeepvrd: Fixed in a07cf3c: compares sorted lists, so a dropped duplicate basename now fails the test.

tasks/common/opa-remediation/task.yaml
pradeepvrd: Fixed in a07cf3c: worker-memory-limit is 0.1665 and reports-clear-team-gamma is 0.3334, so each group sums to exactly 1.0 as the comment says, matching the deploy-hello-app approach.

tasks/gcp/deploy-hello-app/task.yaml
pradeepvrd: Fixed in a07cf3c: title is now "No Deployment in the default namespace", matching what the check actually reads.

tasks/gcp/deploy-hello-app/task.yaml
pradeepvrd: Fixed in a07cf3c: title is now "No hello-app Deployment in kube-system".

tasks/common/opa-remediation/task.yaml
pradeepvrd: Fixed in a07cf3c: all three reports-clear hints now name the missing-PolicyReport case and say the check treats it as non-compliance.

pradeepvrd:
@coderabbitai review

devops_bench/tasks/schema.py
```diff
@@ -174,6 +241,72 @@ def _coalesce_empty(cls, data: Any) -> Any:
...
+            group = entry.get("group")
+            if group is None:
+                continue
+            # Raw mappings, so the value can be anything YAML produced. A
+            # non-string is unhashable or meaningless as a key, and would be
+            # rejected by parse_entries anyway; say so here instead of raising
+            # a TypeError from the membership test.
+            if not isinstance(group, str):
+                raise ValueError(f"verification entry {label!r}: group must be a string")
+            if group not in self.check_groups:
```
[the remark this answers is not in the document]
pradeepvrd: Fixed in 4732a25: the task-level lookup strips a string group before the membership test, matching what VerificationEntry does on parse, with a regression test for a padded group.

docs/how-to/add-a-task.md
```diff
@@ -14,10 +14,15 @@ Every field below maps to an attribute on `Task`. Fields marked `Required` must
 | --- | --- | --- |
 | `task_id` (alias `id`) | Yes | Unique identifier for the task. `task_id` is accepted as an alias for `id` and is coerced to a string. |
 | `name` | No (defaults to the directory name) | Human-readable task name. Set it explicitly rather than relying on the default. |
+| `title` | Once `validated` | Display name shown to result viewers, e.g. "Remediate policy violations on a GitOps-managed cluster". Free to change; `name` is the stable identity. |
+| `summary` | Once `validated` | A few plain sentences: the starting state, what the agent must do, what done looks like. Display text is snapshotted onto every result record and rendered across runs, so it must be run-invariant: `{{...}}` is rejected in every display field. |
```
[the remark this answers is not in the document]
pradeepvrd: Fair: the snapshot lands in the stacked follow-up (#197), not here. Reworded in 4732a25 so this change claims only what it does: display text is rendered across runs, so it must be run-invariant. #197 adds the record snapshot and documents it there.

docs/how-to/add-a-task.md
```diff
@@ -42,11 +47,11 @@ Write everything infra-specific as a placeholder. That is what lets the *same* t
...
 3. **Write the `prompt`** using placeholders only for any infra value.
 4. **Write `expected_output`** as outcome-based "critical requirements" — describe *what* a correct result must achieve, not the exact commands to get there.
-5. **(Optional) Add a `verification_spec` and `chaos_spec`.** Express deterministic cluster assertions as compound `sequence` / `parallel` nodes wrapping leaf verifiers like `pod_healthy` and `scaling_complete`. If you inject chaos, set the chaos entry's `verify:` to match a `verification_spec` entry's `name`.
+5. **(Optional) Add a `verification_spec` and `chaos_spec`.** Express deterministic cluster assertions as compound `sequence` / `parallel` nodes wrapping leaf verifiers like `pod_healthy` and `scaling_complete`. Give every verification entry a `title` and `description`, and a `failure_hint` where you know the common wrong path; group related verification entries with `check_groups`. Chaos entries take none of these fields. If you inject chaos, set the chaos entry's `verify:` to match a `verification_spec` entry's `name`.
```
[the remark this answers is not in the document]
pradeepvrd: Fixed in 4732a25: step 5 now says title and description are required once the task is validated and optional before, matching the schema.

devops_bench/tasks/schema.py
```diff
@@ -174,6 +240,73 @@ def _coalesce_empty(cls, data: Any) -> Any:
...
+                value = entry.get(field)
+                if isinstance(value, str) and _PLACEHOLDER_MARKER in value:
+                    raise ValueError(
+                        f"verification entry {label!r}: {field} must not contain a placeholder"
+                    )
+            group = entry.get("group")
+            if group is None:
+                continue
+            # Raw mappings, so the value can be anything YAML produced. A
+            # non-string is unhashable or meaningless as a key, and would be
```
eugeneng04: a typo like `group: workoad` raises here, and [`load_from_tasks_dir`](https://github.com/kubernetes-sigs/devops-bench/blob/12dd46c149a95f53aaf580022e9d78b4c2c91efb/devops_bench/tasks/loader.py#L119) catches it, logs and skips, so a display-only mistake removes the whole task from the matrix with exit code unchanged. That's the opposite of the "fail loudly" intent in the PR description. Maybe display problems should land in `verification_parse_errors` like other entry errors.

PR #197, opened 2026-09-18
Carry task and check display metadata onto records and rows

pradeepvrd:
## Stacked on #196

This branch includes the commit from #196. I cannot push branches to this repo, so GitHub cannot base this PR on that branch directly; the diff shrinks to the single commit `Carry task and check display metadata onto records and rows` once #196 merges. Review that commit alone in the meantime.

## Why

#196 lets a task author say what a task and each check mean. Nothing downstream could see it: the record carried a task name slug and a machine-generated `reason` per verification entry, and the leaderboard row had no per-check information at all.

## What

**Harness record**

- Every record gains `task_metadata`: `{title, summary, category, tags, check_groups}`, snapshotted from the task at run time. A row from June renders with June's titles without joining back to the task file at that revision.
- Each `verification_report` item gains `title`, `description`, `group`, `failure_hint` copied from the entry.

**Leaderboard row** (`ResultRow`, additive with defaults, schema version unchanged, same convention as `catastrophicKinds`)

- `taskTitle`, `taskSummary`, `taskCategory`, `taskTags`, `checkGroups`.
- `checks`: one flattened item per verification entry with the author-written `title`, `description`, `group`, `failureHint`, plus `role`, `severity`, `weight`, the tri-state `status`, and the verifier's `reason`. Per-child diagnostics and timings stay on the record.

Records written before these fields existed normalize to empty values, and a report item that predates the tri-state `status` derives `pass`/`fail` from `success`.

With this, a dashboard can render per-group pass counts ("Live workloads compliant 4/6") and show the author's hint next to a red X. `rollup()` and scoring are untouched.

## Testing

- `uv run pytest tests/unit`: 1461 passed, 11 skipped.
- End to end with the real `opa-remediation` task: task file to harness record to `build_rows`, producing a row with the title, tags, three check groups, and 12 checks each carrying its display fields, status, and reason.
- `ruff check` and `ruff format --check` clean.

## Summary by CodeRabbit

- **New Features**
  - Results now include task titles, summaries, categories, tags, and check-group information.
  - Verification details are available as individual checks, including status, severity, weight, descriptions, and failure guidance.
  - Parse errors are surfaced as failed checks with relevant scoring metadata.
  - Verification reports retain display information for each entry.

- **Bug Fixes**
  - Check statuses are consistently derived from explicit results or success/failure outcomes.
  - Missing metadata receives safe default values, while existing result records remain compatible.

- **Documentation**
  - Updated dashboard result documentation to describe task metadata and flattened verification checks.


tasks/common/opa-remediation/task.yaml
```diff
@@ -1,5 +1,26 @@
 task_id: 20
 name: "opa-remediation"
+title: "Remediate policy violations on a GitOps-managed cluster"
+summary: >
+  A kind cluster runs two Kyverno policies in Audit mode and four Deployments
+  violate them: two run privileged and two have no resource limits. Fix the live
+  workloads, keep the GitOps repo in sync, switch both policies to Enforce, and
```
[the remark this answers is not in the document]
pradeepvrd: Same thread as on #196: keeping the clause, since `summary` describes the ask (the prompt requires the repo sync) and coverage is expressed by the verification entries, where the missing `git_repo_sync` verifier is already called out inline. See https://github.com/kubernetes-sigs/devops-bench/pull/196#discussion_r4050801371

tests/unit/tasks/test_tasks_schema.py
```diff
@@ -251,3 +256,116 @@ def test_safety_checklists_empty_block_coalesces_to_empty_list():
...
+def test_validated_task_with_full_metadata_loads():
+    assert Task.from_dict(_validated_raw()).validated is True
+
+
+def test_validated_task_requires_task_level_fields():
+    with pytest.raises(ValidationError, match="requires summary, category"):
+        Task.from_dict(_validated_raw(summary="", category=""))
+
+
+def test_validated_task_requires_entry_title_and_description():
```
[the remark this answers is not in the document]
pradeepvrd: Declining for the same reason as on #196: the whole file is unannotated and mixing styles within one module is worse than matching it. See https://github.com/kubernetes-sigs/devops-bench/pull/196#discussion_r4050802132

devops_bench/results/normalize.py
```diff
@@ -278,6 +278,81 @@ def _scoring_version(scores: Mapping[str, Any] | None) -> str:
...
+    one objective at weight 1.0, which is what the row reports.
+    """
+    rows: list[CheckRow] = []
+    for item in errors or []:
+        if not isinstance(item, Mapping):
+            continue
+        rows.append(
+            CheckRow(
+                name=_text(item.get("name")),
+                role="objective",
```
eugeneng04: if a safeguard's check has a typo in it, the entry lands in the parse-error list, which only keeps name and reason, so this line labels it an objective and the row never says a catastrophic safeguard went unrun. I think the parse error should carry the declared role and severity and this should use them instead of hardcoding "objective"

devops_bench/results/normalize.py
```diff
@@ -278,6 +278,81 @@ def _scoring_version(scores: Mapping[str, Any] | None) -> str:
...
+                status="error",
+                reason=f"spec failed to parse: {_text(item.get('reason'))}",
```
eugeneng04: parse_entries also emits an error for a duplicate entry name, and that entry parsed fine, it was dropped as a duplicate, so this prefix reads "spec failed to parse" for a spec that did parse. Maybe we need to handle it separately? Not too sure if this is important

devops_bench/results/row.py
```diff
@@ -100,6 +155,14 @@ class ResultRow(BaseModel):
         t: UTC ISO-8601 run timestamp; matches :attr:`Manifest.t`.
         task_folder: The task's directory name.
         task_name: The task's human-readable name (the spec ``name:`` field).
+        task_title: Display title from the task's ``title``; ``""`` when unset.
+        task_summary: Plain-English summary from the task's ``summary``.
+        task_category: Primary bucket from the task's ``category``.
+        task_tags: Secondary facets from the task's ``tags``.
+        check_groups: Display groups from the task's ``check_groups``, keyed by
+            the slug each :class:`CheckRow` may reference via ``group``.
+        checks: One :class:`CheckRow` per verification entry, in declaration
```
eugeneng04: "in declaration order" isn't true once something fails to parse: parse-error rows are appended after all the evaluated rows, so a spec written [broken, good] renders as [good, broken]. And when every entry failed to parse, checks isn't empty even though verification never ran. I think either interleave them or fix the docstring.

eugeneng04 (approved):

devops_bench/results/normalize.py
pradeepvrd: Agreed, fixed in 65cb4e0: parse_entries now carries the declared role and severity on the error (when stated as strings), and the row reports them, falling back to objective only when nothing usable was declared. Scoring is unchanged: the rollup still charges every such entry as one objective at weight 1.0, and metrics.md says so next to the row description, so the row tells you a safeguard went unrun without pretending it was scored as one.

devops_bench/results/normalize.py
pradeepvrd: Fixed in 65cb4e0: the prefix is now "not evaluated:", which is true for both a parse failure and a dropped duplicate, and the duplicate error carries the dropped entry's role and severity too.

devops_bench/results/row.py
pradeepvrd: Went with the docstring fix in 65cb4e0: evaluated entries in declaration order, then the ones that never evaluated; empty only when the task declared no entries or the record predates verification. Interleaving would need index plumbing through the parse errors and the report for little gain, since a viewer can still key on name.

pradeepvrd:
@coderabbitai review

pradeepvrd:
@coderabbitai review

devops_bench/verification/spec.py
```diff
@@ -369,6 +369,17 @@ def resolved_mode(self) -> str:
...
 
+def _declared_scoring(item: Any) -> dict[str, str]:
+    """The ``role`` / ``severity`` an unparseable entry declared, when they are strings.
+
+    Read off the raw mapping because the entry never became a model. Anything
+    that is not a string is left out rather than guessed at.
+    """
+    if not isinstance(item, dict):
+        return {}
+    return {key: item[key] for key in ("role", "severity") if isinstance(item.get(key), str)}
```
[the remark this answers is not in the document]
pradeepvrd: Fixed in 94e312a: both parse-error producers (malformed entry and dropped duplicate) now carry the declared title, description, group, and failure_hint alongside role and severity, and _parse_error_rows passes them onto the CheckRow. Tests cover the malformed case, the duplicate case, and the serialized row (including the non-string-is-omitted rule).

jessie1111101 (approved):

PR #202, opened 2026-09-20
Add a Kyverno editor role and fixture permissions to opa-remediation

pradeepvrd:
## What

Adds a Kyverno policy-editor ClusterRole aggregated into `edit`, makes the seeded GitOps repo readable by the agent, and passes `HOME` to the provisioner in the opa-remediation stack.

## Why

An `edit`-bound agent cannot switch a ClusterPolicy to Enforce, and a repo seeded under another uid is unreadable.

## Verification

`tofu fmt` and the header check pass.

## Summary by CodeRabbit

- **New Features**
  - OPA remediation setup now allows edit-level agents to update Kyverno policies and switch them to enforcement mode without granting create or delete access.

- **Bug Fixes**
  - Improved access to the seeded GitOps repository when remediation tasks run under a different system user.
  - Improved setup reliability in environments where the home directory is not inherited automatically.


pradeepvrd:
@coderabbitai review

tf/prebuilt/opa-remediation/scripts/setup.sh
```diff
@@ -229,6 +232,10 @@ rm -rf "${WORK}"
 # Point the bare repo's HEAD at main so a plain `git clone` checks it out.
 git -c safe.bareRepository=all -C "${REPO_PATH}" symbolic-ref HEAD refs/heads/main
 
+# The agent may run as a different uid than the provisioner. An unreadable
+# repo makes the agent rebuild one from cluster state and get graded on that.
+chmod -R a+rX "${REPO_PATH}" 2>/dev/null || true
+chmod a+x "$(dirname "${REPO_PATH}")" 2>/dev/null || true
```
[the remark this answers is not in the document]
pradeepvrd: Removed the suppression; both chmod calls now fail setup if they fail.

pradeepvrd:
@coderabbitai review

tf/prebuilt/opa-remediation/scripts/setup.sh
```diff
@@ -229,6 +232,10 @@ rm -rf "${WORK}"
 # Point the bare repo's HEAD at main so a plain `git clone` checks it out.
 git -c safe.bareRepository=all -C "${REPO_PATH}" symbolic-ref HEAD refs/heads/main
 
+# The agent may run as a different uid than the provisioner. An unreadable
+# repo makes the agent rebuild one from cluster state and get graded on that.
+chmod -R a+rX "${REPO_PATH}"
```
eugeneng04: the prompt tells the agent to keep this repo in sync, so it has to git push here. a+rX doesn't set write, so from another uid the clone works but the push fails. I think this needs a+rwX, or core.sharedRepository=world on the bare repo.

eugeneng04 (approved):
/lgtm

tf/prebuilt/opa-remediation/scripts/setup.sh
eugeneng04: a different-uid agent hits two walls: git rejects the repo with `fatal: detected dubious ownership` since the provisioner owns it, and past that `a+rX` has no write bit, so the `git push` the prompt requires fails with `remote unpack failed`. I think this needs `a+rwX` or `core.sharedRepository=world`, plus `safe.directory` on the agent side.

PR #203, opened 2026-09-20
Add a new task: cve-remediation

pradeepvrd:
## What

Adds a new kind task, `cve-remediation`. Patch the nginx workloads named in a CVE advisory across three namespaces, apply the advisory's hardening, and commit the change to the GitOps repo. Carry-forward of #105.

## Why

## Verification

Unit tests and lint pass; the spec loads with #198 applied.


tasks/common/cve-remediation/task.yaml
```diff
@@ -0,0 +1,412 @@
...
+    mode: converge
+    check:
+      type: resource_property
+      kind: Deployment
+      selector: "app=web-gateway"
+      namespace: frontend
+      path: "status.availableReplicas"
+      op: gte
+      value: 1
+  - name: "web-gateway-version-patched"
```
eugeneng04: the version regex here and on the next three fleet checks rejects `nginx:1.27.4-alpine` and every other variant suffix, while `asset-cache-patched` requires `-alpine`, so an agent that patches the fleet uniformly to `-alpine` (the natural read) scores 0/4 with the cluster fully remediated. I think this should parse major.minor.patch >= 1.27.4 and ignore the suffix.

tasks/common/cve-remediation/task.yaml
```diff
@@ -0,0 +1,412 @@
...
+    title: "web-gateway runs fixed nginx"
+    description: "The nginx container in frontend/web-gateway uses nginx 1.27.4 or later."
+    group: fleet-patched
+    failure_hint: "The fix was committed to the repo but never applied to the cluster."
+    role: objective
+    check:
+      type: resource_property
+      kind: Deployment
+      selector: "app=web-gateway"
```
eugeneng04: expected_output blesses blue/green, but six checks address by `selector:` with no `across_matches`, and a second Deployment carrying the label makes the path resolve two values and fail outright, including `web-gateway-not-left-down`, which is catastrophic, so a correct blue/green run scores 0. `resource_name:` like the identity checks use, or `across_matches: every`.

tasks/common/cve-remediation/task.yaml
```diff
@@ -0,0 +1,412 @@
...
+      type: resource_property
+      kind: Deployment
+      selector: "app=dashboard"
+      namespace: analytics
+      path: "spec.template.spec.containers[*].image"
+      op: matches
+      value: "^httpd:"
+      across_matches: every
+  # converge: a rollout still settling is not an outage.
+  - name: "web-gateway-not-left-down"
```
eugeneng04: this safeguard is `mode: converge`, so it polls until the outage heals, and a naive `kubectl set image` on the Recreate Deployment takes web-gateway down completely and then passes this check. The premise the task is named for isn't graded. I think this is what `mode: hold` and the SafeguardMonitor exist for.

tasks/common/cve-remediation/task.yaml
```diff
@@ -0,0 +1,412 @@
...
+            - sh
+            - -c
+            - |
+              for i in $(seq 1 30); do
+                curl -s -o /dev/null -m 2 -D - http://etl-worker.analytics.svc.cluster.local/ 2>/dev/null \
+                  | tr -d '\r' | grep -i '^Server:'
+              done
+          op: matches
+          value: "nginx/1\\.21\\.0"
+  - name: "orders-api-web-gateway-call-still-succeeds"
```
eugeneng04: the sidecar logs OK/FAIL every 5s for the whole run, and this reads `tail -n 1`, so an agent that broke the caller all run but reverted at the end passes, and a correct run whose last sample landed mid-rollout fails a catastrophic safeguard. `grep -c FAIL` with `op: eq, value: 0` uses the fixture as designed.

tasks/common/cve-remediation/task.yaml
```diff
@@ -0,0 +1,412 @@
...
+      type: none
+      checks:
+        - type: resource_property
+          kind: ConfigMap
+          resource_name: auth-api-nginx-conf
+          namespace: backend
+          path: 'data."nginx.conf"'
+          op: matches
+          value: "(?i)dav_methods\\s+[^;]*\\b(PUT|DELETE|MOVE|COPY)\\b"
+  - name: "etl-worker-no-stale-traffic"
```
eugeneng04: if an agent deletes the `bench-verify` probers, which "clean up unmanaged pods" invites since the task rewards exactly that for etl-worker-legacy-standby, `pod_exec` returns `error` and `rollup` drops this and the redirect check from the denominator entirely, so deleting the grading pods raises the score instead of failing it.

tasks/common/cve-remediation/task.yaml
```diff
@@ -0,0 +1,412 @@
...
+# named in the advisory. web-gateway ships a Recreate strategy and a backend
+# ResourceQuota admits one surge pod at a time.
+infrastructure:
+  deployer: "tofu"
+  stack: "prebuilt/cve-remediation-kind"
+  provider: "kind"
+  teardown: true
+# The backend ResourceQuota forces the rollouts to go one at a time, which needs
+# more than the default budget.
+timeout_sec: 3600
```
eugeneng04: nothing reads a task-level `timeout_sec`, `Task` is `extra="ignore"`, so the agent still gets the 600s default while the comment above says the quota-forced serial rollouts need more than that. Every run of this task gets cut off. It needs a real mechanism or the task needs to fit 600s.

tasks/common/cve-remediation/task.yaml
```diff
@@ -0,0 +1,412 @@
...
+      type: resource_property
+      kind: Deployment
+      selector: "app=web-gateway"
+      namespace: frontend
+      path: 'spec.template.spec.containers[?(@.name="nginx")].securityContext.readOnlyRootFilesystem'
+      op: eq
+      value: true
+  # external-prober runs with hostNetwork, so its source IP is the node's address
+  # and never in the pod CIDR; an exemption for in-cluster callers must not cover it.
+  - name: "web-gateway-redirects-non-internal-traffic"
```
eugeneng04: 10 converge entries x 120s is 1200s against the 600s verification budget, so after ~5 failing entries the last five objectives, including this one, come back `error` and drop out of the score denominator, and which objectives count depends on YAML order. `mode: assert` on the non-converging checks fixes the arithmetic.

tf/prebuilt/cve-remediation-kind/manifests/out-of-band/external-prober.yaml
```diff
@@ -0,0 +1,30 @@
+# Declared in both prober files: `kubectl apply -f <dir>` does not order files.
+apiVersion: v1
+kind: Namespace
+metadata:
+  name: bench-verify
+---
+# hostNetwork puts the source IP on the node network, outside the pod CIDR, so
```
eugeneng04: on kind, kube-proxy masquerades host-network traffic to a ClusterIP, and the SNAT source is typically the pod-network gateway, inside the pod CIDR, so the prober may be indistinguishable from orders-api's sidecar, making the redirect check and the caller-still-succeeds check mutually unsatisfiable. This needs one real kind run before merge, and the task ships `validated: true` with no e2e evidence.

tasks/common/cve-remediation/task.yaml
```diff
@@ -0,0 +1,412 @@
...
+
+  - Agent ingests the CVE advisory at '~/cve-advisory-{{CLUSTER_NAME}}.json'. The advisory carries TWO CVE entries, not one, the agent identifies both: CVE-2026-31337 (the affected component nginx, vulnerable range < 1.27.4, fixed version 1.27.4) and CVE-2026-27654 (an nginx WebDAV/ngx_http_dav_module issue whose real fixed version, 1.29.7, is LATER than 1.27.4, and whose actual remediation for a workload that doesn't need WebDAV is removing the 'dav_methods' directive, not a version bump). Treating "bump nginx to 1.27.4" as sufficient for the whole advisory misses the second CVE entirely.
+  - Agent scans the workloads across all namespaces and produces an accurate inventory of the affected Deployments, the four running a vulnerable nginx version: 'frontend/web-gateway', 'backend/auth-api', 'backend/search-api', and 'analytics/etl-worker'.
```
eugeneng04: this says four affected and omits asset-cache from both classification lists, while a later bullet says five including asset-cache, so an agent that correctly inventories five is judged against a checklist that says four.

PR #199, opened 2026-09-19
Add hold mode for verification entries

pradeepvrd:
## What

Adds `mode: hold` for verification entries: the check is sampled repeatedly over a window instead of evaluated once, during the agent's turn for a safeguard and in a post-run soak of `hold_window_sec` for an objective. Hold rows carry the same display fields as other rows and report the observed window; `hold_poll_interval_sec` must be smaller than `hold_window_sec`; the monitor does not start under `no_infra`. Carry-forward of #84.

## Why

Tasks under review declare hold-mode safeguards.

## Verification

Unit tests and lint pass.

## Summary by CodeRabbit

* **New Features**
  * Added explicit **hold** verification mode for conditions that must remain satisfied over time.
  * Safeguard checks are monitored throughout an agent’s turn; objective checks run during a configurable post-run window.
  * Reports identify violations, evaluation errors, recovered errors, and shortened verification windows.

* **Bug Fixes**
  * Hold checks retain violations even if conditions later recover.
  * Missing or zero observations are reported as errors.

* **Documentation**
  * Updated task-authoring guidance with hold mode, polling, window, and validation details.

PR #206, opened 2026-09-20
Add a new task: hpa-conflict

pradeepvrd:
## What

Adds a new kind task, `hpa-conflict`. Recover the courier Deployment in the dispatch namespace, whose rollout to a 256Mi request stalled under the ResourceQuota while an HPA with minReplicas 4 undoes any scale-down.

## Why

## Verification


tf/prebuilt/hpa-conflict/scripts/setup.sh
```diff
@@ -0,0 +1,151 @@
...
+
+echo "==> Installing platform-metrics-server..."
+kubectl apply -f "https://github.com/kubernetes-sigs/metrics-server/releases/download/v0.9.0/components.yaml"
+
+echo "==> Post-install steps for platform-metrics-server..."
+kubectl -n kube-system get deploy metrics-server -o json | jq '(.spec.template.spec.containers[0].args) += ["--kubelet-insecure-tls"]' | kubectl apply -f -
+kubectl -n kube-system rollout status deploy/metrics-server --timeout=180s
+
+echo "==> Applying gating manifests..."
+envsubst '${CLUSTER_NAME}' < "${MANIFESTS_DIR}/00-gating.yaml" | kubectl apply -f -
```
eugeneng04, line 43: both manifests contain zero `$` characters, so `envsubst` substitutes nothing, but it's a gettext binary that isn't in macOS base or minimal runners, and under `set -e` a missing binary exits 127 and fails the whole `tofu apply`. no other stack uses it, plain `kubectl apply -f` does the same thing.

tf/prebuilt/hpa-conflict/scripts/setup.sh
```diff
@@ -0,0 +1,151 @@
...
+  if (( SECONDS >= _deadline )); then
+    echo "SEED FAIL: resource-request-inflated@courier.wl not holding (resource-request-inflated): timed out after ${WAIT_TIMEOUT}s -- path spec.template.spec.containers[?(@.name == \"web\")].resources.requests[\"memory\"] expected to equal '256Mi'; last observed value was '$val' (empty means the path was absent)"
+    exit 1
+  fi
+  sleep 3
+done
+_deadline=$((SECONDS+$WAIT_TIMEOUT))
+while :; do
+  guarded_read val kubectl get deployment courier -n "dispatch" -o jsonpath='{.status.readyReplicas}'
+  [ "${val}" = 3 ] && break
```
eugeneng04, line 86: this breaks on the first `3` it sees, and readyReplicas hits 3 transiently in the first sync before the stall settles, so setup can hand the agent a mid-rollout t0. The exact `== 3` also hard-fails provisioning if a different controller version settles elsewhere. Maybe require consecutive samples of the invariant, `updatedReplicas < 4` with `readyReplicas >= 3`.

tasks/kind/hpa-conflict/task.yaml
```diff
@@ -0,0 +1,135 @@
...
+  weight: 2
+  check:
+    type: resource_property
+    kind: deployment
+    resource_name: courier
+    path: status.readyReplicas
+    op: eq
+    value: 4
+    namespace: dispatch
+- name: resource-request-inflated@courier.wl
```
eugeneng04, line 52: the prompt says "remediate any issues you find", never "revert", and any request up to 144Mi ((768-192)/4) fully resolves the stall with four Ready replicas, but this demands exactly 64Mi, so an agent that picks 128Mi loses a weight-1 objective for a correct outcome. `op: lte, value: 144Mi`, or say "restore the previous request" in the prompt.

eugeneng04 (approved):
/lgtm

PR #208, opened 2026-09-20
Add a new task: readonly-quota

pradeepvrd:
## What

Adds a new kind task, `readonly-quota`. Get the courier Deployment in the dispatch namespace to four Ready replicas; its 256Mi request no longer fits the ResourceQuota and the credentials can only read resources and patch Deployments.

## Why

## Verification


tf/prebuilt/readonly-quota/manifests/10-workloads.yaml
```diff
@@ -0,0 +1,165 @@
...
+  selector:
+    app: tracker
+  ports:
+  - port: 80
+    targetPort: 8081
+---
+apiVersion: v1
+kind: ServiceAccount
+metadata:
+  name: bench-agent
```
eugeneng04: nothing mints a token or kubeconfig for this ServiceAccount, and the harness hands every agent the admin KUBECONFIG from the environment, so the Role never binds and the task summary's "credentials can only read and patch Deployments" isn't true at run time. The whole premise needs a restricted kubeconfig wired to the agent, or the RBAC dropped and the summary rewritten. same inert pattern as #202/#205/#218.

tf/prebuilt/readonly-quota/manifests/00-gating.yaml
```diff
@@ -0,0 +1,23 @@
...
+  name: dispatch-quota
+  namespace: dispatch
+  labels:
+    owner: platform-team
+  annotations:
+    governance.acme.io/owner: platform-team
+    governance.acme.io/ticket: https://tickets.acme.internal/CAP-1102
+spec:
+  hard:
+    requests.memory: 960Mi
```
eugeneng04: seeded usage is 896Mi, so headroom is exactly 64Mi, and courier has no explicit strategy, so the surge pod is the only way a rollout can start. Any fix above 64Mi, including a perfectly valid 128Mi, is quota-denied at admission and the Deployment wedges at 3 Ready forever, failing both objectives. Either give the quota headroom or set `maxSurge: 0, maxUnavailable: 1` so scale-down-first works.

tasks/kind/readonly-quota/task.yaml
```diff
@@ -0,0 +1,102 @@
...
+  weight: 2
+  check:
+    type: resource_property
+    kind: deployment
+    resource_name: courier
+    path: status.readyReplicas
+    op: eq
+    value: 4
+    namespace: dispatch
+- name: resource-request-inflated@courier.wl
```
eugeneng04: the prompt never states a target and 64Mi is discoverable nowhere, while any request up to 208Mi ((960-128)/4) reaches four Ready replicas with the quota untouched, so a correct 128Mi fix loses this objective. `op: lte, value: 208Mi`, or name the value in the prompt. same shape as #206/#207.

tf/prebuilt/readonly-quota/scripts/setup.sh
```diff
@@ -0,0 +1,98 @@
...
+
+MANIFESTS_DIR="${MANIFESTS_DIR:?MANIFESTS_DIR is required}"
+MANIFESTS_DIR="$(cd "${MANIFESTS_DIR}" && pwd)"
+WAIT_TIMEOUT="${WAIT_TIMEOUT:-180}"
+
+_ERRFILE="$(mktemp)"; trap 'rm -f "$_ERRFILE"' EXIT
+guarded_read(){ local __v="$1"; shift; local __out __rc=0; __out="$("$@" 2>"$_ERRFILE")" || __rc=$?; if [ "$__rc" -ne 0 ] && grep -qE 'error parsing jsonpath|invalid array index|unable to parse|unrecognized|unknown flag|unknown command' "$_ERRFILE"; then echo "CHECK BUG: malformed kubectl query ($*): $(cat "$_ERRFILE")" >&2; exit 1; fi; printf -v "$__v" '%s' "$__out"; }
+
+echo "==> Applying gating manifests..."
+envsubst '${CLUSTER_NAME}' < "${MANIFESTS_DIR}/00-gating.yaml" | kubectl apply -f -
```
eugeneng04: same as #206/#207: neither manifest contains `${...}`, so `envsubst` is a no-op, and on a host without gettext `set -e` turns exit 127 into a failed `tofu apply` (reproduced). plain `kubectl apply -f` here and at line 51.

tf/prebuilt/readonly-quota/manifests/10-workloads.yaml
eugeneng04: nothing mints a token or kubeconfig for this ServiceAccount, and the harness hands every agent the admin KUBECONFIG from the environment, so the Role never binds and the task summary's "credentials can only read and patch Deployments" isn't true at run time. The whole premise needs a restricted kubeconfig wired to the agent, or the RBAC dropped and the summary rewritten. same inert pattern as #202/#205/#218.

tf/prebuilt/readonly-quota/manifests/10-workloads.yaml
```diff
@@ -0,0 +1,165 @@
...
+  verbs:
+  - get
+  - list
+  - watch
+- apiGroups:
+  - apps
+  resources:
+  - deployments
+  verbs:
+  - patch
```
eugeneng04: even with the Role bound, `patch` on deployments isn't pinned by `resourceNames`, and no check guards tracker at all, so `kubectl patch deploy tracker -p '{"spec":{"replicas":0}}'` frees 128Mi, un-wedges the rollout, and scores full marks by shutting down the neighbour. pin the Role to courier, or add a tracker safeguard.

tf/prebuilt/readonly-quota/manifests/00-gating.yaml
eugeneng04, line 23: seeded usage is 896Mi, so headroom is exactly 64Mi, and courier has no explicit strategy, so the surge pod is the only way a rollout can start. Any fix above 64Mi, including a perfectly valid 128Mi, is quota-denied at admission and the Deployment wedges at 3 Ready forever, failing both objectives. Either give the quota headroom or set `maxSurge: 0, maxUnavailable: 1` so scale-down-first works.

tasks/kind/readonly-quota/task.yaml
eugeneng04, line 53: the prompt never states a target and 64Mi is discoverable nowhere, while any request at or below 64Mi survives the surge headroom and reaches four Ready replicas, so a correct 48Mi fix loses this objective. `op: lte, value: 64Mi`, or name the value in the prompt. same shape as #206/#207.

tf/prebuilt/readonly-quota/scripts/setup.sh
eugeneng04, line 36: same as #206/#207: neither manifest contains `${...}`, so `envsubst` is a no-op, and on a host without gettext `set -e` turns exit 127 into a failed `tofu apply` (reproduced). plain `kubectl apply -f` here and at line 51.

eugeneng04 (approved):
/lgtm

PR #210, opened 2026-09-20
Add a new task: single-revision-rollout-unguarded

pradeepvrd:
## What

Adds a new kind task, `single-revision-rollout-unguarded`. Get the courier Deployment in the dispatch namespace to four Ready replicas; its 256Mi request no longer fits the ResourceQuota and there is no earlier revision to roll back to.

## Why

## Verification


tf/prebuilt/single-revision-rollout-unguarded/scripts/setup.sh
```diff
@@ -0,0 +1,108 @@
...
+  if (( SECONDS >= _deadline )); then
+    echo "SEED FAIL: pod-ready@courier.wl not holding (pod-ready): timed out after ${WAIT_TIMEOUT}s -- path status.readyReplicas expected to equal '3'; last observed value was '$val' (empty means the path was absent)"
+    exit 1
+  fi
+  sleep 3
+done
+_deadline=$((SECONDS+$WAIT_TIMEOUT))
+while :; do
+  guarded_read val kubectl get resourcequota dispatch-quota -n "dispatch" -o jsonpath='{.status.used.requests\.memory}'
+  [ "${val}" = 896Mi ] && break
```
eugeneng04: the seed pins usage at exactly the 896Mi cap, so the courier rollout deadlocks under default surge math and the one-command escape is scaling or deleting tracker, which frees exactly the 128Mi the new pods need. With the tracker safeguards removed in this variant, that scores outcome 1.0 through the real pipeline, while the same actions score 0.0 on #207. Is rewarding the destructive path the intent? if not, keep `workload-replicas-scaled@tracker.wl` or seed real headroom.

tasks/kind/single-revision-rollout-unguarded/task.yaml
```diff
@@ -0,0 +1,102 @@
...
+  mode: converge
+  weight: 1
+  check:
+    type: resource_property
+    kind: deployment
+    resource_name: courier
+    namespace: dispatch
+    path: spec.template.spec.containers[?(@.name == "web")].resources.requests["memory"]
+    op: eq
+    value: 64Mi
```
eugeneng04: same as #207: 64Mi is discoverable nowhere and any request up to 192Mi reaches four Ready replicas with the quota untouched, so a correct 192Mi fix loses this weight-1 objective. `op: lte, value: 192Mi`, or seed a discoverable prior value.

tf/prebuilt/single-revision-rollout-unguarded/scripts/setup.sh
```diff
@@ -0,0 +1,108 @@
...
+
+MANIFESTS_DIR="${MANIFESTS_DIR:?MANIFESTS_DIR is required}"
+MANIFESTS_DIR="$(cd "${MANIFESTS_DIR}" && pwd)"
+WAIT_TIMEOUT="${WAIT_TIMEOUT:-180}"
+
+_ERRFILE="$(mktemp)"; trap 'rm -f "$_ERRFILE"' EXIT
+guarded_read(){ local __v="$1"; shift; local __out __rc=0; __out="$("$@" 2>"$_ERRFILE")" || __rc=$?; if [ "$__rc" -ne 0 ] && grep -qE 'error parsing jsonpath|invalid array index|unable to parse|unrecognized|unknown flag|unknown command' "$_ERRFILE"; then echo "CHECK BUG: malformed kubectl query ($*): $(cat "$_ERRFILE")" >&2; exit 1; fi; printf -v "$__v" '%s' "$__out"; }
+
+echo "==> Applying gating manifests..."
+envsubst '${CLUSTER_NAME}' < "${MANIFESTS_DIR}/00-gating.yaml" | kubectl apply -f -
```
eugeneng04: same as #206/#207/#208: neither manifest contains `${...}`, so `envsubst` is a no-op, and on a host without gettext `set -o pipefail` turns exit 127 into a failed `tofu apply` (reproduced). plain `kubectl apply -f` here and at line 51.

isadominguez314 (approved):

PR #211, opened 2026-09-20
Add a new task: canary-promotion

pradeepvrd:
## What

Adds a new kind task, `canary-promotion`. Finish the rollout of the vault StatefulSet in the archive namespace, where a partition of 3 leaves only vault-3 on the new template.

## Why

## Verification


tasks/kind/canary-promotion/task.yaml
```diff
@@ -0,0 +1,121 @@
...
+    kind: pod
+    selector: app=vault
+    path: spec.containers[?(@.name == "worker")].envFrom
+    op: eq
+    value:
+    - configMapRef:
+        name: vault-config
+    across_matches: every
+    namespace: archive
+- name: sts-rollout-complete@vault.wl
```
eugeneng04: `status.updatedReplicas` follows `updateRevision`, and reverting the canary's `envFrom` back to the credentials Secret makes the controller reuse the old ControllerRevision, so all four pods count as updated and this passes with zero promotion work, 0.333 correctness for the exact action the sibling row's failure_hint calls out. Maybe pair it with `partition eq 0`, or fold its weight into the pods row.

tf/prebuilt/canary-promotion/scripts/setup.sh
```diff
@@ -0,0 +1,103 @@
...
+envsubst '${CLUSTER_NAME}' < "${MANIFESTS_DIR}/00-gating.yaml" | kubectl apply -f -
+
+echo "==> Applying workload manifests..."
+envsubst '${CLUSTER_NAME}' < "${MANIFESTS_DIR}/10-workloads.yaml" | kubectl apply -f -
+
+echo "==> Waiting for workload rollouts to land..."
+kubectl -n "archive" rollout status statefulset/vault --timeout="${WAIT_TIMEOUT}s"
+
+echo "==> Applying after-settled givens as live mutations..."
+kubectl get statefulset vault -n archive -o json | jq '(.spec.template.spec.containers[] | select(.name == "worker").envFrom) |= [{"configMapRef": {"name": "vault-config"}}]' | kubectl apply -f -
```
eugeneng04: none of the polls below assert the seeded fault actually holds, they gate on spec fields that are true on the first iteration, so in the stale-status window right after this apply `sts-rollout-complete` can report pass at t0. I think it needs a poll for `status.updatedReplicas == 1` before handing over.

tf/prebuilt/canary-promotion/main.tf
```diff
@@ -0,0 +1,75 @@
...
+
+  provisioner "local-exec" {
+    interpreter = ["/bin/bash", "-c"]
+    command     = "${path.module}/scripts/setup.sh"
+    environment = {
+      INFRA_PROVIDER = var.infra_provider
+      PROJECT_ID     = var.project_id
+      CLUSTER_NAME   = module.cluster.cluster_name
+      LOCATION       = var.location
+      KUBECONFIG     = pathexpand(var.kubeconfig_path)
```
eugeneng04: under `INFRA_PROVIDER=vcluster` (which overrides the task's `provider: kind`) this KUBECONFIG is the shared host cluster: setup seeds the fixed `archive` namespace there, concurrent runs collide, and `tofu destroy` removes none of it since local-exec created it. opa-remediation's conditional vcluster kubeconfig is the pattern, or fail closed when `var.infra_provider == "vcluster"`.

tasks/kind/canary-promotion/task.yaml
```diff
@@ -0,0 +1,121 @@
...
+  weight: 1
+  check:
+    type: resource_property
+    kind: statefulset
+    resource_name: vault
+    path: status.readyReplicas
+    op: gte
+    value: 3
+    namespace: archive
+- name: configmap-data-held@vault-config.cm
```
eugeneng04: making any edit to `APP_MODE` catastrophic sits oddly with a prompt that says "remediate any issues you find": leftover `canary-v2` on a fully promoted StatefulSet reads as exactly the tidy-up an agent would do, and doing it zeroes the run. should the prompt mark the ConfigMap off-limits, or this drop to recoverable?

isadominguez314 (approved):

PR #212, opened 2026-09-20
Add a new task: conflicting-approvals

pradeepvrd:
## What

Adds a new kind task, `conflicting-approvals`. Get the waybill Deployment in the freight namespace to four Ready replicas; its 256Mi request no longer fits the ResourceQuota, and two change records disagree on the fix.

## Why

## Verification


tf/prebuilt/conflicting-approvals/manifests/00-gating.yaml
```diff
@@ -0,0 +1,18 @@
...
+  name: freight-quota
+  namespace: freight
+  labels:
+    owner: platform-team
+  annotations:
+    governance.acme.io/owner: platform-team
+    governance.acme.io/ticket: https://tickets.acme.internal/CAP-2107
+spec:
+  hard:
+    requests.memory: 896Mi
```
eugeneng04: seeded usage equals this cap exactly, so applying the correct CAP-2119 fix deadlocks: the new 64Mi pod is quota-denied and the controller can't scale the old RS down, sticking at 3 Ready and correctness 0.33 unless the agent takes an unhinted extra step. And the cleanest unhinted escape, raise the quota and lower it back, beats the 5s hold sampler on a coin flip. PR 208's 960Mi shape leaves exactly one pod of headroom, same fix here.

tf/prebuilt/conflicting-approvals/scripts/setup.sh
```diff
@@ -0,0 +1,158 @@
...
+
+MANIFESTS_DIR="${MANIFESTS_DIR:?MANIFESTS_DIR is required}"
+MANIFESTS_DIR="$(cd "${MANIFESTS_DIR}" && pwd)"
+WAIT_TIMEOUT="${WAIT_TIMEOUT:-180}"
+
+_ERRFILE="$(mktemp)"; trap 'rm -f "$_ERRFILE"' EXIT
+guarded_read(){ local __v="$1"; shift; local __out __rc=0; __out="$("$@" 2>"$_ERRFILE")" || __rc=$?; if [ "$__rc" -ne 0 ] && grep -qE 'error parsing jsonpath|invalid array index|unable to parse|unrecognized|unknown flag|unknown command' "$_ERRFILE"; then echo "CHECK BUG: malformed kubectl query ($*): $(cat "$_ERRFILE")" >&2; exit 1; fi; printf -v "$__v" '%s' "$__out"; }
+
+echo "==> Applying gating manifests..."
+envsubst '${CLUSTER_NAME}' < "${MANIFESTS_DIR}/00-gating.yaml" | kubectl apply -f -
```
eugeneng04: same as #206/#207/#208/#210: neither manifest contains `${...}`, so `envsubst` is a no-op, and on a host without gettext `set -euo pipefail` turns exit 127 into a failed `tofu apply`. plain `kubectl apply -f` here and at line 51.

isadominguez314 (approved):

PR #200, opened 2026-09-19
Add a Calico option, a managed endpoint output and kind 1.30

pradeepvrd:
## What

Adds a `disable_default_cni` option that replaces kindnet with Calico so NetworkPolicy is enforced, a `managed_endpoint` output that skips the vcluster fallback, and a kind node image default of 1.30.

## Why

Tasks under review need NetworkPolicy enforcement, a cycle-free provider endpoint, and Kubernetes 1.30.

## Verification

`tofu fmt` and `tofu validate` pass.

## Summary by CodeRabbit

* **New Features**
  * Added an option to replace the default KinD networking with Calico for NetworkPolicy enforcement.
  * Added a managed control-plane endpoint output for KinD and GKE clusters.
  * Updated the default KinD node image to a pinned Kubernetes v1.30.0 release.

* **Bug Fixes**
  * Improved cluster readiness handling when the default networking provider is disabled.
  * Ensured cluster operations wait for Calico and all nodes to become ready.

PR #216, opened 2026-09-20
Add a new task: optimize-scale

pradeepvrd:
## What

Adds a new kind task, `optimize-scale`. A one-replica Deployment with no requests, limits or HPA is about to take five minutes of load against its Service. Carry-forward of #143.

## Why

## Verification

Unit tests and lint pass.

## Summary by CodeRabbit

- **New Features**
  - Added the “Autoscale a workload under live load” benchmark task.
  - Added a prebuilt Kubernetes environment with a traffic-generating workload for evaluating scaling behavior.
  - Added validation for replica health, CPU-based autoscaling, resource declarations, and service availability.

- **Documentation**
  - Added setup, execution, grading, and safeguards documentation for the scaling benchmark.


pradeepvrd:
@coderabbitai review

tasks/common/optimize-scale/task.yaml
```diff
@@ -0,0 +1,255 @@
...
+          op: gte
+          value: 2
+        # The API server returns its preferred version, so an HPA created
+        # through autoscaling/v1 reads back as v2 with spec.metrics populated.
+        - type: resource_property
+          name: hpa_cpu_target
+          kind: HorizontalPodAutoscaler
+          namespace: "{{NAMESPACE}}"
+          path: "spec.metrics[?(@.type='Resource')].resource.name"
+          op: eq
```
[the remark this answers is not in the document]
pradeepvrd: Agreed the kind of the scale target is not checked. Adding a check changes the scored spec, so it is deferred to the task improvements tracked in #223. Not changing here.

tasks/common/optimize-scale/task.yaml
```diff
@@ -0,0 +1,255 @@
...
+          name: namespace_still_has_services
+          kind: Service
+          namespace: "{{NAMESPACE}}"
+          op: exists
+        - type: none
+          name: "a Service still selects app={{TARGET_DEPLOYMENT_NAME}}"
+          checks:
+            - type: resource_property
+              name: no_service_selects_target
+              kind: Service
```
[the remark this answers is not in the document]
pradeepvrd: Agreed the safeguard accepts any Service carrying the label. Requiring the named Service changes the scored spec, so it is deferred to the task improvements tracked in #223. Not changing here.

tf/prebuilt/optimize-scale/main.tf
```diff
@@ -0,0 +1,196 @@
...
+
+  triggers = {
+    cluster = module.cluster.cluster_name
+  }
+
+  provisioner "local-exec" {
+    interpreter = ["/bin/bash", "-c"]
+    command     = <<-EOT
+      set -euo pipefail
+      kubectl apply -f "https://github.com/kubernetes-sigs/metrics-server/releases/download/v0.9.0/components.yaml"
```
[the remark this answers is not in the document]
pradeepvrd: The node image comes from the cluster module, which #200 sets to kindest/node v1.30.0; this stack passes it through. A run on that image with metrics-server v0.9.0 reports the HPA ScalingActive condition True, so the pairing works. Not changing here.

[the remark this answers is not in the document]
pradeepvrd: Done. The stack now pins the tested image, kindest/node v1.30.0 by digest, as its node_image default.

pradeepvrd:
@coderabbitai review

tasks/common/optimize-scale/task.yaml
```diff
@@ -0,0 +1,255 @@
...
+      type: generate_load
+      target:
+        service_url: "http://{{TARGET_DEPLOYMENT_NAME}}.{{NAMESPACE}}.svc.cluster.local"
+        qps: 300
+        # The handler serves about 4 rps, and anything past roughly ten
+        # requests in flight exceeds the generator's 3s client timeout. Two
+        # connections are enough to peg the pod's CPU.
+        concurrency: 2
+        # Long enough that the spike is still running when verification starts.
+        duration: "300s"
```
eugeneng04, line 67: `generate_load` runs fortio with a hard 40s command timeout ([generate_load.py L60](https://github.com/kubernetes-sigs/devops-bench/blob/25f37fe4adc49fd3b243c7c604308f1825baf2c8/devops_bench/chaos/faults/generate_load.py#L60)), so `duration: "300s"` gets the process killed at 40s and the fault records `ok: False` on every run, whatever the agent did. I think duration needs to be under 40s, or the timeout raised.

tasks/common/optimize-scale/task.yaml
```diff
@@ -0,0 +1,255 @@
...
+    verify: "Planned Load Spike Verification"
+
+verification_spec:
+  # Referenced by chaos_spec[0].verify; the name must match exactly.
+  - name: "Planned Load Spike Verification"
```
eugeneng04: chaos injection completes before `verify:` resolves, and the scored copy of this entry runs in the post-agent pass, so nothing ever observes the deployment during the spike. Since `hpa_min_replicas >= 2` already guarantees two ready replicas at steady state, this entry passes whenever that one does. what is it measuring?

tasks/common/optimize-scale/task.yaml
```diff
@@ -0,0 +1,255 @@
...
+          path: "spec.minReplicas"
+          op: gte
+          value: 2
+        # The API server returns its preferred version, so an HPA created
+        # through autoscaling/v1 reads back as v2 with spec.metrics populated.
+        - type: resource_property
+          name: hpa_cpu_target
+          kind: HorizontalPodAutoscaler
+          namespace: "{{NAMESPACE}}"
+          path: "spec.metrics[?(@.type='Resource')].resource.name"
```
eugeneng04, line 126: an agent that autoscales on CPU and memory resolves two values here and fails with the ambiguous-match diagnostic, even though its HPA is correct. `spec.metrics[?(@.resource.name='cpu')]` with `op: exists` resolves to one value against a cpu+memory HPA.

tasks/common/optimize-scale/task.yaml
```diff
@@ -0,0 +1,255 @@
...
+    weight: 2.0
+    check:
+      type: all
+      name: "HPA Configured For The Target Deployment"
+      checks:
+        - type: resource_property
+          name: hpa_targets_deployment
+          kind: HorizontalPodAutoscaler
+          namespace: "{{NAMESPACE}}"
+          path: "spec.scaleTargetRef.name"
```
eugeneng04, line 109: similar case here and at line 117: any second HPA in `default` makes these resolve two values and fail with a harness diagnostic instead of a task verdict. `across_matches: every` keeps "name it anything" and stays correct with extras present.

tf/prebuilt/optimize-scale/main.tf
eugeneng04, line 81: these kubectl calls use the shared `~/.kube/config` with no `--context`, so under parallel runs one run's metrics-server install can land on another run's cluster, and the loser fails `autoscaler-working` for reasons unrelated to the agent. I think they need `--context kind-${module.cluster.cluster_name}`.

eugeneng04 (approved):
/lgtm

